PR #17, opened 2017-05-03
update kubectl apply doc for required resource name

supereagle:
Update kubectl.yaml for kubernetes/kubernetes#44607

@pwittrock  PTAL.

### DIFF
--- a/gen-kubectldocs/generators/v1_5/kubectl.yaml
+++ b/gen-kubectldocs/generators/v1_5/kubectl.yaml
@@ -241,12 +241,12 @@ toplevelcommandgroups:
   - maincommand:
       name: apply
       synopsis: Apply a configuration to a resource by filename or stdin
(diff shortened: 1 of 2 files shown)


pwittrock:
@supereagle Actually, I don't understand why the new doc says the name must be provided.  What does that mean?

supereagle:
@pwittrock `kubectl apply` will check the resource name and fail if it is missing. Some users may wan to use generateName instead of resource name in `kubectl apply`, so remind users to avoid this wrong usage by improving the doc. The detailed discussion can be referred in kubernetes/kubernetes#44501 and kubernetes/kubernetes#44527.

supereagle:
@pwittrock  Are you clear about the purpose of this PR now?

pwittrock:
Yup.  Thanks.

supereagle:
OK. FYI, I think the PR #18 to revert this change can be closed.

Issue #33, opened 2017-11-22
Missing name field in Pod example in docs


edisonxiang:
**What happened**:
https://kubernetes.io/docs/api-reference/v1.8/#pod-v1-core shows an example of a Pod but the container does not have a name which is a required field (I get an error from kubectl with this example).

**Anything else we need to know?**:  linked with [this issue](https://github.com/kubernetes/kubernetes/issues/55838).

**Environment**:
Kubernetes version (use kubectl version): head

adnavare:
Hi @edisonxiang : Is this issue still open? Are you submitting a new commit?

PR #34, opened 2017-11-22
Add Container Name into Pod Example

edisonxiang:
**what this PR does / why we need it**:

[pod-v1-core example](https://kubernetes.io/docs/api-reference/v1.8/#pod-v1-core) should have a container name so that the users can copy and run the example directly.

**Fixes** #33

**Release note:**

```release-note
None
```


edisonxiang:
Hey @pwittrock, if you have time , could you take a look at this PR?
The ci is failed, I think whether we need to configure or remove the Travis ci service.

PR #42, opened 2018-01-24
Replace regular characters with HTML entities.

steveperry-53:
Addresses [kubernetes/kubernetes issue 57362](https://github.com/kubernetes/kubernetes/issues/57362).

[Related PR in kubernetes/website](https://github.com/kubernetes/website/pull/7038)

The [Kubernetes API reference docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.9/) are not rendering angle brackets in API element descriptions. So when a description includes a placeholder in angle brackets, the entire placeholder is missing from the rendered HTML.

For example, in the currently published reference docs, under [Toleration](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.9/#toleration-v1-core), the description is missing some place holders that are in angle brackets.

The description should say this:

```
The pod this Toleration is attached to tolerates any taint that matches the triple <key,value,effect> using the matching operator <operator>.
```

But instead, is says this:

```
The pod this Toleration is attached to tolerates any taint that matches the triple using the matching operator .
```

This PR adjusts the doc generation code so that descriptions use HTML entities instead of actual angle brackets.

The Definition struct now has an additional field: DescriptionWithEntities.

### DIFF
--- a/gen-apidocs/generators/api/config.go
+++ b/gen-apidocs/generators/api/config.go
@@ -28,6 +28,7 @@ import (
 	"sort"
 	"strings"
 	"unicode"
+	"html"
 
 	"github.com/go-openapi/loads"
 )
@@ -89,7 +90,7 @@ func NewConfig() *Config {
 
 	// Initialize all of the operations
 	config.Definitions = GetDefinitions(specs)
-
+	
 	if *UseTags {
 		// Initialize the config and ToC from the tags on definitions
 		config.genConfigFromTags(specs)
@@ -105,6 +106,11 @@ func NewConfig() *Config {
 
 	// Get the map of operations appearing in the open-api spec keyed by id
 	config.InitOperations(specs)
+
+
+	// In the descriptions, replace unicode escape sequences with HTML entities.
+	config.createDescriptionsWithEntities()
+
 	config.CleanUp()
 
 	// Prune anything that shouldn't be in the ToC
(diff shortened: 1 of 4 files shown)


steveperry-53:
@misterikkit

steveperry-53:
@chenopis Can you take a quick look at this.

chenopis:
@steveperry-53 The code looks fine. Are you asking me about the Travis error?

Also, are there any other escape sequences that need to be converted besides `<` and `>`?

gen-apidocs/generators/api/config.go
```diff
@@ -557,3 +560,19 @@ func doScaleIdHack(version, name, match string) (string, string) {
...
+
+	// The OpenAPI spec has escape sequences like \u003c. When the spec is unmarshaled,
+	// the escape sequences get converted to ordinary characters. For example,
+	// \u003c gets converted to a regular < character. But we can't use  regular <
+	// and > characters in our HTML document. This function replaces the regular
+	// < and > characters with HTML entities.
+	
+	for _, definition := range config.Definitions.GetAllDefinitions() {
+		d := definition.Description()
+		d = strings.Replace(d, "<", "&lt;", -1)
```
misterikkit: Would html.EscapeString be more appropriate/robust here?

steveperry-53: Yes thanks. Changed to html.EscapeString.

PR #43, opened 2018-01-30
Reduce the number of definitions not found.

steveperry-53:
When we generate the Kubernetes API reference docs, we get several errors of the form "No Definition found for &lt;operationID&gt;. This PR makes changes that reduce the number of definitions not found.

This PR removes some lines from the doScaleIdHack function. Those lines applied to the OpenAPI spec for Kubernetes 1.6, but they don't apply to versions after 1.6.

### DIFF
--- a/gen-apidocs/generators/api/config.go
+++ b/gen-apidocs/generators/api/config.go
@@ -90,7 +90,7 @@ func NewConfig() *Config {
 
 	// Initialize all of the operations
 	config.Definitions = GetDefinitions(specs)
-	
+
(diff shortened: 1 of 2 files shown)


heckj:
If it works, it’s good by me - I haven’t read the surrounding code in sufficient detail to really understand what’s happening with those methods, and probably won’t for at least a week right now due to work constraints, so please don’t hold this up on my review account. 

There are other changes, including renaming some dummy links and such in there - those are intentional, yeah?

PR #53, opened 2018-07-17
Rework API reference generator

tengqm:
This PR reworks the generator for API references:

- Generalize the writer and reorganize packages and modules;
- Add new generator for producing HTML result directly;
- Fixed swagger definition that missed SchedulingClass definition.

With this revised version:

- We fix the problem of example code display which no longer occupy 45%
  of the screen;
- The dependency on node JS;
- The improper use of Go Template is gone;
- The dependency on a docker image which team doesn't own is gone.

The current "brodocs" way of reference generation is still there. We
can remove it later when we are satisfied with this new "native"
generator.

To use the old "brodocs" generator, do `make brodocsapi` instead of
`make api`.


tengqm:
The generated doc can be seen here: https://deploy-preview-9536--kubernetes-io-master-staging.netlify.com/docs/reference/generated/kubernetes-api/v1.11/

PR #56, opened 2018-08-13
Initial OWNERS file

chenopis:
fixes https://github.com/kubernetes-incubator/reference-docs/issues/54

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,0 +1,24 @@
+# Reviewers can /lgtm /approve but not sufficient for auto-merge without an
+# approver
+reviewers:
+- Rajakavitha1
+- stewart-yu
+- xiangpengzhao
+- zhangxiaoyu-zidif
+
+# Approvers have all the ability of reviewers but their /approve makes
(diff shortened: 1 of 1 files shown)


OWNERS
```diff
@@ -0,0 +1,24 @@
+# Reviewers can /lgtm /approve but not sufficient for auto-merge without an
+# approver
+reviewers:
+- Rajakavitha1
```
mistyhacks: Can we not use Github groups?

spiffxp: >We use aliases for groups instead of GitHub Teams, because changes to GitHub Teams are not publicly auditable.

https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md#owners_aliases

That was true the time these files were initially written, github's teams have improved and that may no longer be true, we're planning to revisit that down the line

spiffxp:
I'm going to follow this up with a kubernetes/org issue to get everyone in this file added as a kubernetes-incubator org member

spiffxp:
Manually merging

PR #64, opened 2018-10-11
type fix from propogating to propagating

liyongxin:


xiangpengzhao:
/lgtm
Thanks for the fix!

PR #66, opened 2018-12-12
Add tool-reference templating

kbhawkey:
- add tool-reference template commands to
  generate markdown for tool components.
- kubeadm is still rendered with notitle, but
  could be changed to use tool-reference template.
- tested the generation of all components using the
  update-imported-docs script.
- this is a first pass; further clean-up needed.

### DIFF
--- a/gen-compdocs/generators/doc.go
+++ b/gen-compdocs/generators/doc.go
@@ -24,8 +24,6 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
-	// "time"
-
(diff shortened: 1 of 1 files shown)


kbhawkey:
@tengqm, with these changes, and the tool-reference template from k8s/website, the component docs should generate.

kbhawkey:
- tool-reference.html template was added to kubernetes/website/layout/partials/template
- the generator uses the template instructions to generate uniform tool reference documentation

PR #67, opened 2018-12-29
Update apps versions used in examples

tengqm:
ref: kubernetes/website#9687


alexbrand:
Does it make sense to update other resources as well? DaemonSets and ReplicaSets seem to be using old versions in the examples as well.

PR #77, opened 2019-05-08
Escape special chars in long descriptions

tengqm:


DanyC97:
@tengqm if you don't mind asking, to which issue is this PR related to ? trying to see where the special char were used in the long description

tengqm:
@DanyC97 From slack channel:

yamamoto [6:57 PM]
how is `website/content/en/docs/reference/setup-tools/kubeadm/generated` generated?

yamamoto [6:57 PM]
i guess it needs to escape `<`

yamamoto [6:58 PM]
eg `specified as ":",` in https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-join/#synopsis

tengqm:
> Also, I doubt that the 1.14 tool component docs (kube-proxy, kube-scheduler, etc.) are using this new rendering -- I still see the numbered TOC at the top.

This change will impact all components regarding their "long" description. I'm not sure I understand the "numbered TOC" thing.

> Does the description produced by the script contain the substituted characters or is the problem with the Markdown generation by Hugo template?

This change detects whether there are `<` or `>` in the long description and substitute them if so. We are not using Hugo template at this stage. The docs generator produces markdown files for Hugo to render. So it is possible to fix this at Hugo (downstream) side as well.

gen-compdocs/generators/doc.go
```diff
@@ -79,15 +79,18 @@ func GenMarkdownCustom(cmd *cobra.Command, w io.Writer, linkHandler func(string)
...
 		}
 
 		if _, err := fmt.Fprintf(w, "%s\n\n", "{{% capture synopsis %}}"); err != nil {
 			return err
 		}
-		
+
+		// Escape any '<', '>' characters found in the long description
+		long = strings.Replace(long, "<", "&lt;", -1)
+		long = strings.Replace(long, ">", "&gt;", -1)
```
yamt: does `>` need an escape? i thought escaping `<` was enough.

gen-compdocs/generators/doc.go
```diff
@@ -122,35 +127,31 @@ func GenMarkdownCustom(cmd *cobra.Command, w io.Writer, linkHandler func(string)
...
 			}
 		}
-
-	}	
-
+	}
 
 	/* Examples */
 	if len(cmd.Example) > 0 {
-	if with_title {
```
yamt: these indent fixes are unrelated to the main change?

tengqm: Yes. The indentation problem hurts the readability when I tried to find out the location to insert this new logic, so I fixed them together.

gen-compdocs/generators/doc.go
tengqm: I think the best practice would be escaping both.

gen-compdocs/generators/doc.go
kbhawkey: Thanks for fixing the indentation.

gen-compdocs/generators/doc.go
kbhawkey: I think there are a couple of ways to fix this, depending on what is generated in this script:
1) calling a single method on the entire string (long description) to escape ALL chars that need escaping -- is there a Go library to do this?
OR
2) leaving the string alone here and test if we can clean up the string (long description) using Hugo.

kbhawkey:
>> Also, I doubt that the 1.14 tool component docs (kube-proxy, kube-scheduler, etc.) are using this new rendering -- I still see the numbered TOC at the top.
> 
> This change will impact all components regarding their "long" description. I'm not sure I understand the "numbered TOC" thing.
> 

Right, @tengqm. If I look at the version 1.14 of the docs on k8s.io/docs/reference, the component docs generated are not generated with this version of the script (there is a table a contents at the top of each page (or numbered list) -- not an issue. 
It was my impression, that the `kubeadm` docs take a different path in the script. I will re-read the changes; I may have misunderstand the `kubeadm` processing. The character escaping problem seems like a recent problem, but, will affect all components.

> This change detects whether there are `<` or `>` in the long description and substitute them if so. We are not using Hugo template at this stage. The docs generator produces markdown files for Hugo to render. So it is possible to fix this at Hugo (downstream) side as well.

Right. I got it. See my other comment about fixing here in a generic way, or trying to push the clean up of the chars to the Hugo processing.

gen-compdocs/generators/doc.go
kbhawkey: hello @tengqm. Does this update need further review? What about adding a single call,
fmt.Println(html.EscapeString(s)), to take care of all the html chars?

tengqm: Before conversion, the input data are scattered in many different elements in the specification. This means we don't have a HTML to perform this operation. After the conversion, the whole doc is a HTML string, which is not meant to be escaped. The only thing we can do, as I see it, is somewhere during the conversion. We inject the customization logic wherever appropriate.

Did I capture what you said?

kbhawkey: Yes, this change seems fine, though I did not get a chance to verify. LGTM. Thanks.

PR #78, opened 2019-05-09
Print timestamp for API reference doc

tengqm:


gen-apidocs/generators/html.go
```diff
@@ -584,8 +585,10 @@ func (h *HTMLWriter) generateHTML(navContent string) {
 		}
 	}
 
+	const layout = "Jan 2, 2006 at 03:04pm (MST)"
 	fmt.Fprintf(html, "%s", navContent)
 	fmt.Fprintf(html, "<BR/><DIV class=\"copyright\">%s</DIV></DIV>\n", h.TOC.Copyright)
+	fmt.Fprintf(html, "<BR/><DIV>Generated at: %s</DIV></DIV>\n", time.Now().Format(layout))
```
kbhawkey: Great @tengqm. This may be a long shot, but would it be possible to call a Git API/Webhook to publish the commit ID when the final html is published in k8s/website?
I realize that HTML is generated here, and not Markdown. The other doc pages generated from Markdown publish a commit ref in a footer?

tengqm: Em, not sure how to extract commit ID from Go. Will explore.

Other docs don't yet have this logic.

kbhawkey: OK. It would be a great to publish the updated date. Perhaps the commit could be added as a post publishing step from a script in k8s/website.

tengqm: I did revisited the whole flow again in attempt to figure out the right "timestamp" value or git commit ID to print. Let's take 1.14 version as an example. The current fix is about printing the date when the API docs are generated, i.e. the date the generator was executed. So .. it is not very useful if you skipped `make updateapispec` (https://github.com/kubernetes-incubator/reference-docs/blob/master/Makefile#L60). Here are some alternatives:

- have the `api` target depend on the `updateapispec` target, so that each time you do `make api`, you are copying the lates swagger.json from the upstream repo. I'm doubting if this is helpful.
- explicitly document `make updateapispec` step (if we haven't done so), so the release maester knows the difference between running this step and skipping it.

The next question is about timestamp. I'm now a little bit confused about the intent of printing it. What are we trying to achieve with the timstamps. Depending on the goals, there are a few timestamp values to choose from:

- The HEAD of the specified branch (e.g. release 1.14) of the upstream repo. This helps us track the latest swagger.json we are using. For example, in release-1.14 branch, we have 1.14.1, 1.14.2, ... tags.
- The last modified date of the Swagger JSON file. Considering that the spec is not changing very frequently, this may be more useful a timestamp if we are focusing on the Open API itself rather than the whole upstream repo. The shortcoming is that we still need to map the datetime value to K8S release version (a difficult math problem I believe).
- The HEAD of the reference-docs repo. I believe we need branches to map the tool version to the upstream release as well. At the end of the day, the HEAD should be given on a per-branch basis. With this commit ID, we would be able to track problems in the generator.
- You can think of other timestamp or git commit IDs as well...

kbhawkey: Thanks for looking into this, @tengqm . 
Including the generated API reference into a k8s/website markdown page would add a `PR merge timestamp` to the reference. I'll give this some more thought. 
Still, there is missing metadata about the version of the spec used to generate the ref and the actual version of the release. What happens if someone makes an upstream change and wants to publish the reference between release cycles? I am guessing an updated spec needs to be copied over from k8s master once the change has been committed, followed by publishing a new API ref version. How is this update tracked or the publishing changes linked together?

Ideas for version info that publish as part of the API reference:
- spec version
- release version
- copyright in the left nav, states 2016 and links to k8s/k8s. Update to 2016 - 2019?

tengqm: Good suggestion.

tengqm:
Preview:

image

kbhawkey:
@tengqm , looks great! Did you remove the copyright from the bottom left nav?

tengqm:
@kbhawkey Yes I did.

Makefile
```diff
@@ -1,5 +1,5 @@
 WEBROOT=~/src/github.com/kubernetes/website
-K8SROOT=~/src/github.com/kubernetes/kubernetes
+K8SROOT=~/k8s/src/k8s.io/kubernetes
```
kbhawkey: I was able to build the API ref with the original K8SROOT setting, though, I had to change to the k8s.io/kubernetes to build the kubectl reference (for example, src/k8s.io/kubernetes).

kbhawkey:
Great! With these changes, are there additional steps required to build the API ref?
LGTM.

tengqm:
It looks like we missed some API endpoints due to hard coded paths: 

https://github.com/kubernetes-incubator/reference-docs/blob/master/gen-apidocs/generators/api/config.go#L271

This function needs a change so that some other API endpoints will be included in the build.

PR #80, opened 2019-06-29
update Makefile for kubectl, api ref builds

kbhawkey:
- update Makefile to correct destination paths.
- fixed error parsing _generated_rollout.md.


tengqm:
This is nice.

PR #84, opened 2019-09-12
Update API reference generator for 1.16

tengqm:
There are some changes to the top-level API resources in 1.16. One of the weird change is about the newly added `TokenRequest` resource definition where the sole operation defined for it is called `CreateCoreV1NamespacedServiceAccountToken`. Why we are creating a resource type named `Potato` and adding an operation called `CookMushroom`? Sigh.


tengqm:
@simplytunde This PR updates the swagger.json from the upstream repo. The version is 1.16.0.

tengqm:
Kicking this in to unblock ref doc generation.

PR #86, opened 2019-09-12
cleanup commands

kbhawkey:
- cleanup kube-apiserver, kubelet commands
- update .travis.yml


tengqm:
@kbhawkey You may need to update .travis for the KUBE_VERSION used.

kbhawkey:
> @kbhawkey You may need to update .travis for the KUBE_VERSION used.

Still investigating what happened to the tar file after v1.14.

PR #88, opened 2019-09-21
Docbook output for apidocs

feloy:
This PR adds a backend to gen-apidocs, to output the documentation in the DocBook format, then transorms the DocBook source in PDF and HTML.

To create the doc in DocBook format:
```
$ make dbapi
```

To create the PDF (US letter) from DocBook format:
```
$ make dbapi-pdf-letter
```

To create the PDF (A4) from DocBook format:
```
$ make dbapi-pdf-a4
```

An overview of HTML and PDF outtputs are available here;

http://k8s-v1-16-api.surge.sh/index.html

http://k8s-v1-16-api.surge.sh/index-a4.pdf (6.5 MB, 1932 pages)
http://k8s-v1-16-api.surge.sh/index-letter.pdf (6.4 MB, 2026 pages)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,6 +63,33 @@ updateapispec:
 api: cleanapi
 	go run gen-apidocs/main.go --config-dir=gen-apidocs/generators --munge-groups=false
 
+mdapi: cleanapi
+	go run gen-apidocs/main.go --backend=brodocs --config-dir=gen-apidocs/generators --munge-groups=false
+
+# DocBook / HTML / PDF generation
+dbapi: cleanapi
+	go run gen-apidocs/main.go --backend=docbook --config-dir=gen-apidocs/generators --munge-groups=false
+
+dbapi-html: gen-apidocs/generators/build/index.xml
+	(cd gen-apidocs/generators/build && \
+	 mkdir -p html && \
+	 cd html && \
+	 xsltproc /usr/share/xml/docbook/stylesheet/docbook-xsl/html/chunk.xsl ../index.xml)
+
+dbapi-pdf-a4: gen-apidocs/generators/build/index.xml
+	(cd gen-apidocs/generators/build && \
+	 mkdir -p pdf-a4 && \
+	 cd pdf-a4 && \
+	 xsltproc --stringparam paper.type A4 -o index-a4.fo ../../../../xsl/api.xsl ../index.xml && \
+	 fop -pdf index-a4.pdf -fo index-a4.fo)
+
+dbapi-pdf-letter: gen-apidocs/generators/build/index.xml
+	(cd gen-apidocs/generators/build && \
+	 mkdir -p pdf-letter && \
+	 cd pdf-letter && \
+	 xsltproc --stringparam paper.type USletter -o index-letter.fo ../../../../xsl/api.xsl ../index.xml && \
+	 fop -pdf index-letter.pdf -fo index-letter.fo)
+
 # NOTE: The following "sudo" may go away when we remove docker based api doc generator
 cleanapi:
 	sudo rm -rf $(shell pwd)/gen-apidocs/generators/build
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ dbapi-html: gen-apidocs/generators/build/index.xml
 	(cd gen-apidocs/generators/build && \
 	 mkdir -p html && \
 	 cd html && \
-	 xsltproc /usr/share/xml/docbook/stylesheet/docbook-xsl/html/chunk.xsl ../index.xml)
+     xsltproc ../../../../xsl/api-html.xsl ../index.xml)
 
 dbapi-pdf-a4: gen-apidocs/generators/build/index.xml
 	(cd gen-apidocs/generators/build && \
--- a/Makefile
+++ b/Makefile
@@ -80,14 +80,14 @@ dbapi-pdf-a4: gen-apidocs/generators/build/index.xml
 	(cd gen-apidocs/generators/build && \
 	 mkdir -p pdf-a4 && \
 	 cd pdf-a4 && \
-	 xsltproc --stringparam paper.type A4 -o index-a4.fo ../../../../xsl/api.xsl ../index.xml && \
+	 xsltproc --stringparam fop1.extensions 1 --stringparam paper.type A4 -o index-a4.fo ../../../../xsl/api.xsl ../index.xml && \
 	 fop -pdf index-a4.pdf -fo index-a4.fo)
 
 dbapi-pdf-letter: gen-apidocs/generators/build/index.xml
 	(cd gen-apidocs/generators/build && \
 	 mkdir -p pdf-letter && \
 	 cd pdf-letter && \
-	 xsltproc --stringparam paper.type USletter -o index-letter.fo ../../../../xsl/api.xsl ../index.xml && \
+	 xsltproc --stringparam fop1.extensions 1 --stringparam paper.type USletter -o index-letter.fo ../../../../xsl/api.xsl ../index.xml && \
 	 fop -pdf index-letter.pdf -fo index-letter.fo)
 
 # NOTE: The following "sudo" may go away when we remove docker based api doc generator
--- a/gen-apidocs/generators/docbook.go
+++ b/gen-apidocs/generators/docbook.go
@@ -1,0 +1,491 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package generators
+
+import (
+	"fmt"
+	"html"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/kubernetes-incubator/reference-docs/gen-apidocs/generators/api"
+)
+
+type DocbookTOCItem struct {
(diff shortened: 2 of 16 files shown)


kbhawkey:
helllo @feloy. This is interesting. What do you plan to do with the output?
/cc @tengqm

feloy:
> helllo @feloy. This is interesting. What do you plan to do with the output?
> /cc @tengqm

The chunked HTML output would be used to replace the one page HTML at https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/

The PDF / ePub outputs would be downloadable from the kubernetes.io website.

I would also work on some "short" version of the PDF version, without Old Versions and probably without Operations parts, and with additional text and examples, intended for beginners.

kbhawkey:
> The chunked HTML output would be used to replace the one page HTML at https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/

@feloy, my initial thought about [the proposed output](http://k8s-v1-16-api.surge.sh/index.html) is that the pages, representing the API, would be difficult to navigate as a website. The current API page also has some navigation and presentation quirks and could use updating.

feloy:
> > The chunked HTML output would be used to replace the one page HTML at https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/
> 
> @feloy, my initial thought about [the proposed output](http://k8s-v1-16-api.surge.sh/index.html) is that the pages, representing the API, would be difficult to navigate as a website. The current API page also has some navigation and presentation quirks and could use updating.

@kbhawkey yes, I agree that the default HTML output visible at https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/ is not easy to navigate. But by working on the stylesheets transforming DocBook to HTML, it would be possible to get a presentation and look and feel similar to the other kubernretes.io rubrics (for ex. https://kubernetes.io/docs/concepts/) , with the table of contents on the left and short pages on the right.

kbhawkey:
@feloy. Just checked out the styling. Have you thought about a three column flex layout for displaying the API reference? I _think_ limiting the amount of items/folders that require opening in the left nav list is important. There are a number of open source projects with API references that have a main navigation (left side), a single page per category (multiple endpoints), and a right hand navigation to traverse the links within the page (For example: istio.io, agones. Also check out the docsy Hugo theme). For a reference guide, tables work well to display the information, especially for field variables and definitions, request/response codes.

feloy:
@kbhawkey thanks for the suggestion. Here is a new revision with 3 columns, the page ToC being on the right, or at the beginning of the page when the page is not large enough: http://k8s-v1-16-api.surge.sh/index.html

tengqm:
Tried making the PDF version of API doc, got the following error a million times:

```
[ERROR] FOUserAgent - Invalid property value encountered in column-width="proportional-column-width(1)": org.apache.fop.fo.expr.PropertyException: file:/home/tengqm/k8s/src/github.com/kubernetes-incubator/reference-docs/gen-apidocs/generators/build/pdf-letter/index-letter.fo:5803:12844: proportional-column-width() function may only be used when fo:table has table-layout="fixed".; property:'column-width' (See position 5803:13004) <org.apache.fop.fo.expr.PropertyException: file:/home/tengqm/k8s/src/github.com/kubernetes-incubator/reference-docs/gen-apidocs/generators/build/pdf-letter/index-letter.fo:5803:12844: proportional-column-width() function may only be used when fo:table has table-layout="fixed".; property:'column-width'>org.apache.fop.fo.expr.PropertyException: file:/home/tengqm/k8s/src/github.com/kubernetes-incubator/reference-docs/gen-apidocs/generators/build/pdf-letter/index-letter.fo:5803:12844: proportional-column-width() function may only be used when fo:table has table-layout="fixed".; property:'column-width'
```

feloy:
Hi @tengqm you are right, I get the same errors. I added an option to xsltproc to support the FOP processor particularities. Thanks

feloy:
Hi @kbhawkey @tengqm , is there any interest in this PR, or do I close it?

tengqm:
@feloy I'm inclined to merge this once the issues are resolved.

kbhawkey:
Hi @feloy. I will spend some time tomorrow to look at the generated pages and the tools used to generate the pages. It seems as if the content could use additional styling (to look like a reference) and spacing between the sections on the page. I find the output a bit squished together.

feloy:
Thanks,

I think it could be interesting to clearly separate the 2 processes:

1. Create a docbook file from the API spec with gen-apidocs
2. Create an html website and other pdf files from this Docbook file.

As a matter of fact, 
- the Docbook file can be used in a lot of different cases, with different stylesheets,
- creating a website that fits in the k8s.io/docs website will need some information that are not available in this repo (for ex CSS stylesheets), but are (most probably) in the Kubernetes/website repo.

kbhawkey:
> 
>     * the Docbook file can be used in a lot of different cases, with different stylesheets,
> 
>     * creating a website that fits in the k8s.io/docs website will need some information that are not available in this repo (for ex CSS stylesheets), but are (most probably) in the Kubernetes/website repo.

Hi @feloy. A few more questions/observations:
- Does the target, `mdapi` replace `api` in the Makefile?
- I am in the process of reading docbook.go.
- Static includes? Is there a way to automate these files or will the docs team need to update/maintain when the API changes?
- Could the publishing tools run in a container? What additional tools are needed to build (libxslt (xsltproc), fop + java, docbook) the new targets?
- I have not looked at DocBook for a while. What are the drawbacks of using DocBook?
Would it be easier to use golang templates to generate markdown?
- I am in favor of reusing the styling from the k8s/website -- less to maintain. The final targets could generate Markdown or HTML and then copy over.
- The reference documentation should fit with the rest of the documentation. Granted the k8s/website style may change (new theme) and depends upon Hugo to generate the HTML.
- Could DocBook use a template (xslt stylesheet) to create output that is relevant to the API reference components (for example: table of fields (this is most useful), (Write, Read, Create, Patch, List, Watch, Replace, Delete) ops, response codes, examples, (path, query, body)parameters)? Once the main components are created, the HTML page could use styling/custom styling from k8s/website to style the individual components. Right now it feels as if the text is just running on the page. I would like to see more page structure.

tengqm:
@feloy Any comments to the feedback?

feloy:
Hi @kbhawkey @tengqm I'm not completely satisfied with the implementation I've done here (and some points of the feedback highlight the problems). I'm exploring other possibilities in parallel. Thanks for your feedback

feloy:
> > ```
> > * the Docbook file can be used in a lot of different cases, with different stylesheets,
> > 
> > * creating a website that fits in the k8s.io/docs website will need some information that are not available in this repo (for ex CSS stylesheets), but are (most probably) in the Kubernetes/website repo.
> > ```
> 
> Hi @feloy. A few more questions/observations:
> 
> * Does the target, `mdapi` replace `api` in the Makefile?
> * I am in the process of reading docbook.go.
> * Static includes? Is there a way to automate these files or will the docs team need to update/maintain when the API changes?
> * Could the publishing tools run in a container? What additional tools are needed to build (libxslt (xsltproc), fop + java, docbook) the new targets?
> * I have not looked at DocBook for a while. What are the drawbacks of using DocBook?

The primary advantages I see to Docbook are for the pdf (and epub) outputs. I do not see an important one for the html output, as Hugo translates Markdown into html.

>   Would it be easier to use golang templates to generate markdown?
> * I am in favor of reusing the styling from the k8s/website -- less to maintain. The final targets could generate Markdown or HTML and then copy over.

The best would be to create Markdown, directly handled by Hugo, so the style would be the same of other pages.


> * The reference documentation should fit with the rest of the documentation. Granted the k8s/website style may change (new theme) and depends upon Hugo to generate the HTML.
> * Could DocBook use a template (xslt stylesheet) to create output that is relevant to the API reference components (for example: table of fields (this is most useful), (Write, Read, Create, Patch, List, Watch, Replace, Delete) ops, response codes, examples, (path, query, body)parameters)? Once the main components are created, the HTML page could use styling/custom styling from k8s/website to style the individual components. Right now it feels as if the text is just running on the page. I would like to see more page structure.

kbhawkey:
> > > * the Docbook file can be used in a lot of different cases, with different stylesheets,

> The primary advantages I see to Docbook are for the pdf (and epub) outputs. I do not see an important one for the html output, as Hugo translates Markdown into html.

Hi @feloy . I see your point. The existing k8s OpenAPI spec could be used to generate multiple output targets html, pdf, epub. Though, I am guessing that your immediate focus is the pdf/epub targets. I think the epub/pdf target should be easy, for anyone, to reproduce. Capturing the build and tool dependencies in a container image seems to make sense to me, unless there is another way that you had in mind.

feloy:
> > > > * the Docbook file can be used in a lot of different cases, with different stylesheets,
> 
> > The primary advantages I see to Docbook are for the pdf (and epub) outputs. I do not see an important one for the html output, as Hugo translates Markdown into html.
> 
> Hi @feloy . I see your point. The existing k8s OpenAPI spec could be used to generate multiple output targets html, pdf, epub. Though, I am guessing that your immediate focus is the pdf/epub targets. I think the epub/pdf target should be easy, for anyone, to reproduce. Capturing the build and tool dependencies in a container image seems to make sense to me, unless there is another way that you had in mind.

Hi @kbhawkey yes, I completely agree with you. I will work on a Docker image to build the PDF/epub outputs, as it is done with the image pwittrock/brodocs for building the resource docs.

feloy:
> Hi @feloy. A few more questions/observations:
> * Static includes? Is there a way to automate these files or will the docs team need to update/maintain when the API changes?

I would like to use a MarkdownToDocbook to insert the static includes into the docbook. I'll have a look on what exists and if the result is satisfying.

feloy:
> Hi @feloy. A few more questions/observations:
> 
> * Does the target, `mdapi` replace `api` in the Makefile?

There is perhaps a misunderstanding about this. I have added the target `mdapi` in the Makefile to test the Markdown backend (md.go). It is not related to my PR, and will remove it.

Why is this existing Markdown output not used to be included in the actual kubernetes.io website with Hugo? Is it planned?

feloy:
I prefer to close this PR. I instead worked on a separate project: https://github.com/feloy/kubernetes-resources-reference

tengqm:
Thank you. @feloy

PR #90, opened 2019-09-26
Add comments to html.go generateHTML

aimeeu:
Add comments detailing where css and js files are located.

I'm learning how API generation is done and had to track down the location of the CSS and JS files. I added comments so others don't have to track down the file locations.

### DIFF
--- a/gen-apidocs/generators/html.go
+++ b/gen-apidocs/generators/html.go
@@ -543,6 +543,12 @@ func (h *HTMLWriter) generateHTML(navContent string) {
 		os.Exit(1)
 	}
 
+	/*
+	Physical location of files referenced in this block:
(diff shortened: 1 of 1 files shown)


gen-apidocs/generators/html.go
```diff
@@ -543,6 +543,12 @@ func (h *HTMLWriter) generateHTML(navContent string) {
 		os.Exit(1)
 	}
 
+	/*
+	Physical location of files referenced in this block:
+	reference-docs/gen-apidocs/generators/static/bootstrap.min.css
+	reference-docs/gen-apidocs/generators/static/font-awesome.min.css
+	reference-docs/gen-apidocs/generators/static/stylesheet.css
+	*/
```
tengqm: Ideally, we should let website host these files.

Issue #97, opened 2019-12-19
bootstrap.min.css (4.3.1), js


kbhawkey:
The updated bootstrap.min.css file is `js` not css: 
https://github.com/kubernetes-sigs/reference-docs/blob/master/gen-apidocs/generators/static/bootstrap.min.css

Solution:
Fix the generation of head in index.html to use the cdn or obtain the css for 4.3.1.
```
<link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
```

PR #95, opened 2019-11-01
Update bootstrap to 4.3.1 to match k8s/website

aimeeu:
Update bootstrap to 4.3.1 to match k8s/website. 
gen-apidocs/generators/html.go uses boostrap.min.js. reference-docs version should match the version used by k8s/website. I recall that there was an issue in the generated docs with buttons not functioning with bootstrap 3.

Signed-off-by: Aimee Ukasick <aimeeu.opensource@gmail.com>

PR #98, opened 2019-12-21
Fix CSS file for API reference

tengqm:
closes: #97


tengqm:
bug introduced in PR #95

kbhawkey:
Hi @tengqm . Thanks for quickly fixing this bug. Ping me if there is a PR to review. I thought self approving is discouraged.

tengqm:
@kbhawkey As I have commented, the bug was introduced in PR #95, which was approved by me. It is kind of cleaning up a mess made accidentally. Yes, I get the idea of democracy, I get the idea of having more eyes reviewing things is better. This repo, like many others, is not hot one. There are not many eyes on it, TBH. You can see that from the history of commits. As one of the owners, I have to take the ownership and sometimes take the informed risks of making mistakes.
As a side note, we are all busy, so we don't have a lot of cycles revisiting the same nimble, trivial thing again and again. There are a lot of more interesting, more important things to do in life.
With that, I'm glad you are watching this repo and you really care about it. Thank you.

Issue #100, opened 2019-12-24
api reference, v1.17, toc, missing definitions


kbhawkey:
```
Could not find definition for resource in TOC: discovery v1alpha1 EndpointSlice.
No Definition found for deleteFlowcontrolApiserverV1alpha1FlowSchema [/apis/flowcontrol.apiserver.k8s.io/v1alpha1/flowschemas/{name}].  
No Definition found for patchFlowcontrolApiserverV1alpha1FlowSchema [/apis/flowcontrol.apiserver.k8s.io/v1alpha1/flowschemas/{name}].  
No Definition found for replaceFlowcontrolApiserverV1alpha1FlowSchema [/apis/flowcontrol.apiserver.k8s.io/v1alpha1/flowschemas/{name}].  
No Definition found for readFlowcontrolApiserverV1alpha1FlowSchema [/apis/flowcontrol.apiserver.k8s.io/v1alpha1/flowschemas/{name}].  
No Definition found for patchFlowcontrolApiserverV1alpha1PriorityLevelConfigurationStatus [/apis/flowcontrol.apiserver.k8s.io/v1alpha1/prioritylevelconfigurations/{name}/status].  
No Definition found for replaceFlowcontrolApiserverV1alpha1PriorityLevelConfigurationStatus [/apis/flowcontrol.apiserver.k8s.io/v1alpha1/prioritylevelconfigurations/{name}/status].  
No Definition found for readFlowcontrolApiserverV1alpha1PriorityLevelConfigurationStatus [/apis/flowcontrol.apiserver.k8s.io/v1alpha1/prioritylevelconfigurations/{name}/status].  
No Definition found for watchFlowcontrolApiserverV1alpha1PriorityLevelConfiguration [/apis/flowcontrol.apiserver.k8s.io/v1alpha1/watch/prioritylevelconfigurations/{name}].  
No Definition found for createFlowcontrolApiserverV1alpha1FlowSchema [/apis/flowcontrol.apiserver.k8s.io/v1alpha1/flowschemas].  
No Definition found for listFlowcontrolApiserverV1alpha1FlowSchema [/apis/flowcontrol.apiserver.k8s.io/v1alpha1/flowschemas].  
No Definition found for deleteFlowcontrolApiserverV1alpha1CollectionFlowSchema [/apis/flowcontrol.apiserver.k8s.io/v1alpha1/flowschemas].  
[...]

PR #101, opened 2019-12-25
Fix generator for 1.17 API reference

tengqm:
Adapt API reference generator to changes made to the upstream OpenAPI definitions, i.e.

- New API group `flowcontrol.apiserver.k8s.io` has been added and the group name is not following the `<group>.k8s.io` convention. Sigh...
- New resources `FlowSchema`, `PriorityLevelConfiguration` have been added to the above group;
- Resource `EndpointSlice` has been promoted to `v1beta1` from `v1alpha1`;
- Resource `CSINode` has been promoted to `v1` (GA) from `v1beta1`.

Since a `release-1.16` branch has been created before this PR. This change is targeting master branch, for 1.17+ versions. The Makefile and the swagger.json file have been updated accordingly. 

Closes: #100

### DIFF
--- a/gen-apidocs/generators/api/definition.go
+++ b/gen-apidocs/generators/api/definition.go
@@ -224,6 +224,10 @@ func (d *Definition) GetOperationGroupName() string {
 	if strings.ToLower(d.Group.String()) == "rbac" {
 		return "RbacAuthorization"
 	}
+	if strings.ToLower(d.Group.String()) == "flowcontrol" {
+		return "FlowcontrolApiserver"
+	}
+
 	return strings.Title(d.Group.String())
 }
 
--- a/gen-apidocs/generators/config.yaml
+++ b/gen-apidocs/generators/config.yaml
@@ -64,7 +64,7 @@ resource_categories:
       version: v1
       group: core
     - name: EndpointSlice
-      version: v1alpha1
+      version: v1beta1
       group: discovery
     - name: Ingress
       version: v1beta1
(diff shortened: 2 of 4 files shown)


tengqm:
@kbhawkey

gen-apidocs/generators/api/definition.go
```diff
@@ -224,6 +224,10 @@ func (d *Definition) GetOperationGroupName() string {
 	if strings.ToLower(d.Group.String()) == "rbac" {
 		return "RbacAuthorization"
 	}
+	if strings.ToLower(d.Group.String()) == "flowcontrol" {
+		return "FlowcontrolApiserver"
```
kbhawkey: @tengqm , why are `flowcontrol` and `rbac` special cases? How is the return string determined?

gen-apidocs/generators/config.yaml
```diff
@@ -180,6 +183,9 @@ resource_categories:
       version: v1
       group: core
       description_note: "These are assigned to <a href=\"#pod-v1-core\">Pods</a> using <a href=\"#persistentvolumeclaim-v1-core\">PersistentVolumeClaims</a>."
+    - name: PriorityLevelConfiguration
+      version: v1alpha1
```
kbhawkey: Would it be possible to auto generate this configuration file?
Why are there only a few `description_note` fields?

gen-apidocs/generators/api/definition.go
kbhawkey: Oh, I missed the detail in your commit note. OK.

kbhawkey:
Changes look fine. I will confirm when I rebuild.
/lgtm

PR #102, opened 2019-12-27
Drop 'brodocs' backend for API reference generation

tengqm:
We cannot afford maintaining two backends for generating the reference
docs. This PR drops the legacy 'brodocs' way of building refernece docs.
We drop it because:

- The brodocs container has to be managed by a specific person so the
  team collaboration model is not scalable.
- The brodocs container has not been maintained for a long period.
- The markdown output from the brodocs container are difficult to tune
  for web presentation -- main use case of the tool.
- The new 'html' generator has been used for several release cycles and
  it is generating satisfactory output for users. It can be considered
  mature.


gen-apidocs/generators/writer.go
```diff
@@ -62,15 +58,7 @@ func GenerateFiles() {
...
-	var writer DocWriter
-	if *Backend == "brodocs" {
-		writer = NewMarkdownWriter(config, copyright, title)
-	} else if *Backend == "go" {
-		writer = NewHTMLWriter(config, copyright, title)
-	} else {
-		panic(fmt.Sprintf("Unknown backend specified: %s", *Backend))
-	}
-
+    writer := NewHTMLWriter(config, copyright, title)
```
kbhawkey: nit: indentation of line 61?

gen-apidocs/generators/util.go
```diff
@@ -84,74 +84,3 @@ func PrintInfo(config *api.Config) {
 	//	}
 	//}
 }
-
```
kbhawkey: Is PrintInfo(), JsonOutputFile(manifest.json) used?

gen-apidocs/generators/writer.go
tengqm: will check

tengqm: indentation should be one TAB according Go convention. I have got TABs autoconverted to 4 spaces in my VIM. Will fix.

gen-apidocs/generators/util.go
tengqm: PrintInfo is still referenced, by JsonOutputFile is not (will fix).

kbhawkey: @tengqm , How do these changes affect the generation of the kubectl command reference?

```
docker run -v $(shell pwd)/gen-kubectldocs/generators/includes:/source -v $(shell pwd)/gen-kubectldocs/generators/build:/build -v $(shell pwd)/gen-kubectldocs/generators/:/manifest pwittrock/brodocs
```
Eventually the tooling/build of the kubectl reference will be updated (I hope to reopen the related kubectl reference generation PR and continue work) too.

tengqm: This change is irrelevant to the gen-kubectldocs. This PR is only related to the API reference generation. I'd like to have the kubectl reference reworked as well.

kbhawkey:
@tengqm , Looks fine. Do you know what version of the k8s API was the last version to be generated from the `brodocs` backend?

tengqm:
> @tengqm , Looks fine. Do you know what version of the k8s API was the last version to be generated from the `brodocs` backend?

The "new" generator was merged in July 2018 (https://github.com/kubernetes-sigs/reference-docs/pull/53). So IIRC, 1.11 was the last version that used `brodocs` for API reference generation.

PR #103, opened 2019-12-27
Restruct API generator

tengqm:
This change is mainly for code maintenance. The current directory layout
is confusing for (new) developers. Static CSS files, template files and
input swagger files are all scattered inside the 'generators' directory.
It is difficult to tell which part serves what purposes. This PR
proposes a new directory layout:

```
  gen-apidocs
    |-- build/              Directory for output files.
    |     `-- includes      Temporary files for generating the output files
    |-- config/             Config/data files.
    |     |-- config.yaml   Main configuration file
    |     |-- swagger.json  The current OpenAPI specification.
    |     |-- sections/     Prebuilt templates for sections in the reference.
    |     `-- examples/     Example files referenced in the docs.
    |-- static/             Files referenced from the generated HTML.
    |     |-- css/          CSS files
    |     |-- js/           JavaScript files
    |     `-- fonts/        FontAwesome font files
    `-- generators/         Go code for generating the reference doc.
          `-- api/          Swagger API parser related code.
```


Makefile
```diff
@@ -58,32 +56,30 @@ copycomp:
...
 cleanapi:
-	sudo rm -rf $(shell pwd)/gen-apidocs/generators/build
-	sudo rm -rf $(shell pwd)/gen-apidocs/generators/includes
+	rm -rf $(shell pwd)/gen-apidocs/build
 
 copyapi:
 	cp $(APISRC)/build/index.html $(APIDST)/index.html
 	cp $(APISRC)/build/navData.js $(APIDST)/navData.js
-	cp $(APISRC)/static/scroll.js $(APIDST)/scroll.js
+	cp $(APISRC)/config/static/scroll.js $(APIDST)/scroll.js
```
kbhawkey: nit: why is `scroll.js` considered configuration?

Makefile
```diff
@@ -58,32 +56,30 @@ copycomp:
 
 # Build api docs
 updateapispec:
-	cp $(K8SROOT)/api/openapi-spec/swagger.json gen-apidocs/generators/openapi-spec/swagger.json
+	cp $(K8SROOT)/api/openapi-spec/swagger.json gen-apidocs/config/swagger.json
```
kbhawkey: @tengqm , If the Open API k8s spec (swagger.json) is copied from k8s/k8s, is it necessary to keep a copy of this file
in this repo? Could the file copy to the temporary build location?

Makefile
```diff
@@ -58,32 +56,30 @@ copycomp:
...
-	cp $(APISRC)/static/fontawesome-webfont.svg $(APIDST)/fonts/fontawesome-webfont.svg
-	cp $(APISRC)/static/fontawesome-webfont.ttf $(APIDST)/fonts/fontawesome-webfont.ttf
-	cp $(APISRC)/static/fontawesome-webfont.woff $(APIDST)/fonts/fontawesome-webfont.woff
-	cp $(APISRC)/static/fontawesome-webfont.woff2 $(APIDST)/fonts/fontawesome-webfont.woff2
+	cp $(APISRC)/config/static/FontAwesome.otf $(APIDST)/fonts/FontAwesome.otf
+	cp $(APISRC)/config/static/fontawesome-webfont.eot $(APIDST)/fonts/fontawesome-webfont.eot
+	cp $(APISRC)/config/static/fontawesome-webfont.svg $(APIDST)/fonts/fontawesome-webfont.svg
+	cp $(APISRC)/config/static/fontawesome-webfont.ttf $(APIDST)/fonts/fontawesome-webfont.ttf
+	cp $(APISRC)/config/static/fontawesome-webfont.woff $(APIDST)/fonts/fontawesome-webfont.woff
+	cp $(APISRC)/config/static/fontawesome-webfont.woff2 $(APIDST)/fonts/fontawesome-webfont.woff2
```
kbhawkey: Are lines 84 - 94 needed?

Makefile
tengqm: em, it is a terminology confusion I think. I don't see a clear answer for the best location of the static files. We can move it out of the `config` subdirectory, but that is still confusing. If the `config` directory name is not good, do you have a better option?

Makefile
tengqm: Ideally, we should avoid caching a copy of the swagger.json file. However, we observed that the upstream swagger.json file may get changed after a release is out. In other words, swagger.json from 1.17.0 may differ from that from 1.17.1. We are tracking a release branch say `release-1.17`, not a specific tag. The implication is that we are shooting at a moving target. Keeping a local copy is a simplification of the process. We effectively only care about versions like 1.16.0, 1.17.0. During the release week, we update the JSON once, and adapt the generator accordingly. After that, during the whole development cycle of the next release, we rarely update the JSON.

kbhawkey: OK. @tengqm , I would like this sorted out so that the spec is automatically updated as part of the release (see my changes to the Makefile, see https://github.com/kubernetes/website/pull/18067.). 
Personally, I think it is important to support a simplified way of generating the references -- one k8s/website "wrapper" build/setup script, one time setup of repos to build, and minimal or no manual script updates for versioning.
I plan on updating the Makefile once the restructuring is completed and will need to fix the command that copies the spec from k8s/k8s (add checks before copying).

Two ideas to address versioning of `swagger.json`.
Copy the release version of the file to this repo from k8s/k8s and:
- copy/rename the file to a versioned filename, for example, (config_dir/swagger.json_1.17). As part of reference generation, this file is copied to <some location>/`swagger.json`. The script/Makefile checks if the versioned file in this repo exists before copying from k8s/k8s.
- identify as a file in a versioned directory.

In the case that there is an error in the API (not a generation error) and the API reference needs to be regenerated, a new, different versioned Open API spec will be used and replace the previously copied version?

Makefile
kbhawkey: Hi @tengqm. Since the files are copied to k8s/website, the directory location is not as important.
Why is this change needed? Just to clean up the directory hierarchy?
The directory names help `document` the contents.
When a new theme is in place, these static files may not be needed, or just some may be needed.
You could leave the files in the static directory or the static files could be placed in a directory structure such as:
```
gen-apidocs
    |-- static/              Directory for static files.
    |    -- css/
    |    -- fonts/
    |    -- js/
```

Makefile
tengqm: Sounds a good idea to me. Adding to your proposal, maybe we should couple the config.yaml file with the corresponding swagger JSON file. The reason, as you may have noticed, is that resources are added and removed (deprecated) from one version to another version. A versioned directory for both files sounds a solution to me.

Makefile
tengqm: Okay, let's try your way and see if it looks better.
Yes, the intent behind this PR is to clean up the directory hierarchy. As a developer, I myself sometimes get confused by the files/directories when trying to fix things. :)

Makefile
```diff
@@ -58,32 +56,27 @@ copycomp:
...
+	mkdir -p $(APIDST)/js
+	cp $(APISRC)/build/navData.js $(APIDST)/js/
+	cp $(APISRC)/static/js/* $(APIDST)/js/
+	# copy stylesheet.css, bootstrap.min.css, font-awesome.min.css
 	mkdir -p $(APIDST)/css
-	cp $(APISRC)/static/stylesheet.css $(APIDST)/css/stylesheet.css
-	cp $(APISRC)/static/bootstrap.min.css $(APIDST)/css/bootstrap.min.css
-	cp $(APISRC)/static/jquery.scrollTo.min.js $(APIDST)/jquery.scrollTo.min.js
-	cp $(APISRC)/static/font-awesome.min.css $(APIDST)/css/font-awesome.min.css
+	cp $(APISRC)/static/css/* $(APIDST)/css/font-awesome.min.css
```
kbhawkey: @tengqm , would you double check line 76. 
```
cp $(APISRC)/static/css/* $(APIDST)/css/
```

Makefile
```diff
@@ -58,32 +56,27 @@ copycomp:
...
+	rm -rf $(shell pwd)/gen-apidocs/build
 
 copyapi:
+	mkdir -p $(APIDST)
 	cp $(APISRC)/build/index.html $(APIDST)/index.html
-	cp $(APISRC)/build/navData.js $(APIDST)/navData.js
-	cp $(APISRC)/static/scroll.js $(APIDST)/scroll.js
+	# copy scroll.js, jquery.scrollTo.min.js and the new navData.js
+	mkdir -p $(APIDST)/js
+	cp $(APISRC)/build/navData.js $(APIDST)/js/
```
kbhawkey: Are you creating a new js sub-directory and reference in the index footer?
The current structure is, for example:
$HOME/website/static/docs/reference/generated/kubernetes-api/v1.17/jquery.scrollTo.min.js
$HOME/website/static/docs/reference/generated/kubernetes-api/v1.17/index.html
$HOME/website/static/docs/reference/generated/kubernetes-api/v1.17/scroll.js
$HOME/website/static/docs/reference/generated/kubernetes-api/v1.17/navData.js

gen-apidocs/generators/api/definition.go
```diff
@@ -278,7 +278,7 @@ func (d *Definition) GetResourceName() string {
 }
 
 func (d *Definition) initExample(config *Config) {
-	path := filepath.Join(*ConfigDir, config.ExampleLocation, d.Name, d.Name + ".yaml")
+	path := filepath.Join(*ConfigDir, "config", config.ExampleLocation, d.Name, d.Name + ".yaml")
```
kbhawkey: @tengqm , The config-dir arg is passed to main(). 
```go run gen-apidocs/main.go --config-dir=gen-apidocs --munge-groups=false```

Though, you pass in the root gen-apidocs directory , not gen-apidocs/config. Nit: The argument could be renamed.
Nit: Could you simplify the path/dir calls.
Create a few static strings for `$APISRC/build`, `$APISRC/build/includes`, `$APISRC/config` and pass in `$APISRC` to main()? Since munge-groups is false (default?) do you need to pass in this argument?

File: Makefile (hunk 5 above)
Comment (kbhawkey) [reply Makefile]: OK. I see you have changed the k8s/website directory structure for the js files in `index.html`.

File: Makefile (hunk 2 above)
Comment (kbhawkey) [reply Makefile]: 👍

File: Makefile (hunk 4 above)
Comment (kbhawkey) [reply Makefile]: Besides the few nits, looks fine 😃 . If time, would suggest updating the commit comment
with the updated directory structure.

Comment (tengqm) [reply Makefile]: Done

File: gen-apidocs/generators/api/definition.go (hunk 6 above)
Comment (tengqm) [reply gen-apidocs/generators/api/definition.go]: Optimizing the variables is something we can do in this PR.

File: gen-apidocs/generators/html.go [hunk 7]
```diff
@@ -571,29 +565,35 @@ func (h *HTMLWriter) generateHTML(navContent string) {
 
 	for _, sec := range h.TOC.Sections {
 		fmt.Printf("Collecting %s ... ", sec.File)
-		content, err := ioutil.ReadFile(filepath.Join(*api.ConfigDir, "includes", sec.File))
+		content, err := ioutil.ReadFile(filepath.Join(api.IncludesDir, sec.File))
 		if err == nil {
 			buf += string(content)
+			fmt.Printf("\033[32mOK\033[0m\n")
```
kbhawkey: nit: Create static strings to represent: `Not Found` and `OK`. Replace literals.

gen-apidocs/generators/html.go
```diff
@@ -163,7 +163,7 @@ func (h *HTMLWriter) WriteDefinition(d *api.Definition) {
 	fmt.Fprintf(f, "<H2 id=\"%s\">%s</H2>\n", linkID, nvg)
 	fmt.Fprintf(f, "<TABLE class=\"col-md-8\">\n<THEAD><TR><TH>Group</TH><TH>Version</TH><TH>Kind</TH></TR></THEAD>\n<TBODY>\n")
 	fmt.Fprintf(f, "<TR><TD><CODE>%s</CODE></TD><TD><CODE>%s</CODE></TD><TD><CODE>%s</CODE></TD></TR>\n",
-	            d.GroupDisplayName(), d.Version, d.Name)
+				d.GroupDisplayName(), d.Version, d.Name)
```
kbhawkey: nit: check that the indentations are correct throughout this file.

gen-apidocs/generators/api/operation.go
```diff
@@ -174,7 +174,7 @@ func (o *Operation) GetGroupVersionKindSub() (string, string, string, string) {
 // initExample reads the example config for an operation
 func (o *Operation) initExample(config *Config) {
 	path := o.Type.Name + ".yaml"
-	path = filepath.Join(*ConfigDir, config.ExampleLocation, o.Definition.Name, path)
+	path = filepath.Join(ConfigDir, config.ExampleLocation, o.Definition.Name, path)
 	path = strings.Replace(path, " ", "_", -1)
```
kbhawkey: nit: This seems confusing, `config.ExampleLocation`.
Instead create a static string to represent the `examples` sub dir under `config`.
Or is there a reason to set `example_location: examples` in config.yaml?
Also:
- See line 103 in config.go
- See line 281, definition.go

kbhawkey: Also, double check the variables for `includes` and `static_includes` directories in `writer.go`.

gen-apidocs/generators/html.go
```diff
@@ -571,29 +565,35 @@ func (h *HTMLWriter) generateHTML(navContent string) {
 
 	for _, sec := range h.TOC.Sections {
 		fmt.Printf("Collecting %s ... ", sec.File)
-		content, err := ioutil.ReadFile(filepath.Join(*api.ConfigDir, "includes", sec.File))
+		content, err := ioutil.ReadFile(filepath.Join(api.IncludesDir, sec.File))
```
kbhawkey: @tengqm , Check where the `config/sections` directory variable is set?

gen-apidocs/generators/html.go
tengqm: Done

gen-apidocs/generators/html.go
tengqm: ... this is really about where TABs are expanded and how they are expanded. sigh.

gen-apidocs/generators/api/operation.go
tengqm: No comment to the `examples` setting. It is legacy code I didn't care to fix.

`static_includes` is retrieved from a dedicated function, `getStaticIncludesDir`, we can easily kill it.

gen-apidocs/generators/html.go
tengqm: This has nothing to do with section files. The so called "sections" are actually the old "static_includes".

kbhawkey: Right. I am trying to track down where the files are read/copied from the new dirs `config/sections` as well as `config/examples`.

kbhawkey: Never mind. Found it.
The changes seem fine now. Have you rebuilt the API reference?

tengqm: No. Haven't done that.

gen-apidocs/generators/html.go
```diff
@@ -602,8 +602,8 @@ func (h *HTMLWriter) generateHTML(navContent string) {
 
 	/*
 	Physical locations of files referenced:
-	reference-docs/gen-apidocs/generators/static/jquery.scrollTo.min.js
-	reference-docs/gen-apidocs/generators/static/scroll.js
+	reference-docs/gen-apidocs/static/js/jquery.scrollTo.min.js
+	reference-docs/gen-apidocs/static/js/scroll.js
```
kbhawkey: @tengqm , a couple more nits, then ready to merge.
- The comment on line 608 is misleading as the API reference does not use bootstrap-*.min.js.
- Line 621, pull out js/bootstrap-4.3.1.min.js. This file is not in the original code or Makefile?
- I will rebuild the ref using the changes once this goes in.

tengqm: We do use the bootstrap JS implicitly, for toggling the embedded examples.

kbhawkey: Where is bootstrap*.js stored?

tengqm: https://github.com/kubernetes/website/blob/master/static/js/bootstrap-4.3.1.min.js

kbhawkey: OK

gen-apidocs/generators/html.go
```diff
@@ -617,16 +617,16 @@ func (h *HTMLWriter) generateHTML(navContent string) {
 	fmt.Fprintf(html, "%s", string(buf))
 	fmt.Fprintf(html, "</DIV>\n</DIV>\n")
 	fmt.Fprintf(html, "<SCRIPT src=\"/js/jquery-3.2.1.min.js\"></SCRIPT>\n")
-	fmt.Fprintf(html, "<SCRIPT src=\"jquery.scrollTo.min.js\"></SCRIPT>\n")
+	fmt.Fprintf(html, "<SCRIPT src=\"js/jquery.scrollTo.min.js\"></SCRIPT>\n")
 	fmt.Fprintf(html, "<SCRIPT src=\"/js/bootstrap-4.3.1.min.js\"></SCRIPT>\n")
```
tengqm: @kbhawkey Note the way bootstra-4.3.1.min.js is referenced. The path has a prefix '/'. That is how we reference the existing script in website

kbhawkey: Changes seem fine now. Agreed?

tengqm: Yes. Per the intent of this PR, it is fine now. There are still many places we can iterate and improve in future PRs.

kbhawkey: 👍

Issue #105, opened 2019-12-29
API ref, navbar text strings


kbhawkey:
**Problem**:
The strings in the API reference left navbar list exceed the fixed display area.

**Solution**:
Possibly increase the width for this component or modify the
overflow-x field to `auto`.
Update stylesheet.css.

```
#sidebar-wrapper {
    display: block;
    height: 100%;
    width: 20%;
    position: fixed;
    z-index: 1;
    top: 0;
    left: 0;
    background-color: whitesmoke;
    border-right: 2px solid slategrey;
    overflow-x: hidden;
    padding-top: 60px;
}
```

tengqm:
Em.. good idea.
I'm also thinking if we should allow the left nav bar to be toggled on/off.

PR #110, opened 2020-01-04
Fix HTML display for API reference

tengqm:
The main content of the generated HTML is somehow "too wide" to show in
the main screen area. The left navigation bar cannot accomodate long
lines. These two issues are fixed in this PR.

Closes: #105


kbhawkey:
I am rebuilding the reference now.

Issue #109, opened 2020-01-04
kubectl: versioned directory, static files


kbhawkey:
The kubectl cmd reference attempts to version the files in the static_includes directory as well as the toc.yaml file. Typically, this structure is manually updated per release, but has not been kept up to date. Periodically the toc.yaml file requires a change for successful doc generation.

- create a script to update the static files in this repo for new releases.
- create a designated branch for new releases (version the files).

kbhawkey:
Plan to write up some notes I have about suggested way to accomplish this task.
Temporarily assigning myself until I publish the notes.
/assign

kbhawkey:
See PR #108 .

Issue #108, opened 2019-12-31
api ref: versioned directory (swagger.json,config.yaml)


kbhawkey:
Create a mechanism to version swagger.json and config.yaml.

Note from PR #103:

```
@kbhawkey 
Two ideas to address versioning of swagger.json.
Copy the release version of the file to this repo from k8s/k8s and:
    - copy/rename the file to a versioned filename, for example, (config_dir/swagger.json_1.17). As part of reference generation, this file is copied to /swagger.json. The script/Makefile checks if the versioned file in this repo exists before copying from k8s/k8s.
    - identify as a file in a versioned directory.

In the case that there is an error in the API (not a generation error) and the API reference needs to be regenerated, a new, different versioned Open API spec will be used and replace the previously copied version?

@tengqm 
Sounds a good idea to me. Adding to your proposal, maybe we should couple the config.yaml file with the corresponding swagger JSON file. The reason, as you may have noticed, is that resources are added and removed (deprecated) from one version to another version. A versioned directory for both files sounds a solution to me.
```

kbhawkey:
Temporarily assigning myself.
/assign

kbhawkey:
I think this repo needs to maintain the files required for generating the versioned docs.
The static files need to be accessible and exist from build to build (assets), not temporary build files.

I am in favor of a more permanent solution:
Create a script to update the static files in this repo for each new release. Run the script and commit the changes as a dependency of the reference docs release process.
Or
Create a designated branch for new releases (version the files).

PR #112, opened 2020-01-05
create version dir targets

kbhawkey:
- create make targets for kubectl-command and api-ref to generate versioned directories and files. I expect the kubectl related targets will change when the kubectl-command gen process changes.
- tested kubectl directory generation and updated the versioned files in this repo.
- See issues #108, #109.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,8 +35,8 @@ cleancli:
 	sudo rm -rf $(shell pwd)/gen-kubectldocs/generators/build
 	sudo rm -rf $(shell pwd)/gen-kubectldocs/generators/manifest.json
 
-cli: createkubectldir cleancli
-	go run gen-kubectldocs/main.go --kubernetes-version v1_$(MINOR_VERSION)
+cli: cleancli
+	go run gen-kubectldocs/main.go --kubernetes-version v$(shell cat "release.tmp")
 	docker run -v $(shell pwd)/gen-kubectldocs/generators/includes:/source -v $(shell pwd)/gen-kubectldocs/generators/build:/build -v $(shell pwd)/gen-kubectldocs/generators/:/manifest pwittrock/brodocs
 
 copycli: cli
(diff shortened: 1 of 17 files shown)


kbhawkey:
The general idea is to setup versioned directories/files for a release (running the update-imported-docs script from k8s/website) and save this state.
The generated directories (and files) need to be committed to this repo.

Makefile
```diff
@@ -1,6 +1,18 @@
...
+# set the following environment variables
+# to match your environment
+#
+# WEBROOT=~/src/github.com/kubernetes/website
+# K8SROOT=~/k8s/src/k8s.io/kubernetes
+# K8S_RELEASE=17
+
+WEBROOT=${K8S_WEBROOT}
+K8SROOT=${K8S_ROOT}
+MINOR_VERSION=${K8S_RELEASE}
```
tengqm: I was expecting that the ${K8S_RELEASE} be a full version string like `1.17`.

kbhawkey: @tengqm , I could change the value to be the `major.minor` version. Currently, the kubectl-command directories follow a pattern such as:
v1_17. Could drop the `v` in the directory name.

kbhawkey: A couple of thoughts:
- kubectl is not versioned in k8s/website
- api ref is versioned in k8s/website, but creates a directory like `v1.17` instead of `v1_17`
- the makefile is more complicated. Could introduce a versioning config that saves two vars:
latest and prev. Use these variables for copying config files from release to release.

kbhawkey:
Still thinking on this. Could move the directory creation (and checks) to a shell script.
targets to setup release dirs:
make getversion
make sepversion
make createkubectldir
make createapidir

tengqm:
I'd vote against making the Makefile a script.

kbhawkey:
@tengqm , the shell script should be run before a release to create the versioned directories and copy over files. The rest of the build targets now use a K8S_RELEASE env variable such as `1.17`.

tengqm:
@kbhawkey Let's kick this in and iterate over. We are never gonna solve all issues in a single PR anyway.

PR #114, opened 2020-01-06
Add selector field to Deployment v1 apps example

oke-py:
Is this right place to modify kubectl example in Kubernetes API reference?

closes https://github.com/kubernetes/website/issues/18477


tengqm:
/lgtm
/approve
Thanks.

PR #120, opened 2020-01-08
correct updateapispec tgt

kbhawkey:
Correcting an error in `updateapispec` target. Retested in k8s/website.


tengqm:
em... it is funny why we need such a change.
/lgtm
/approve

kbhawkey:
> em... it is funny why we need such a change.

Ermm, the script did not work without the changes (tested on Ubuntu) :smiley: .

PR #126, opened 2020-01-10
    Make sure responses from operations are sorted

tengqm:
One observation from committing API reference changes to website repo is that there are many fake changes in the responses section. The response code is not sorted. This behavior is undesired. It makes the output unpredictable thus the changes are unclear from a PR commit.

We can sort the responses by their code easily, actually.

### DIFF
--- a/gen-apidocs/generators/html.go
+++ b/gen-apidocs/generators/html.go
@@ -23,6 +23,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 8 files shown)


gen-apidocs/generators/html.go
```diff
@@ -292,7 +293,11 @@ func (h *HTMLWriter) writeResponseParams(w io.Writer, o *api.Operation) {
 
 	fmt.Fprintf(w, "<H3>Response</H3>\n")
 	fmt.Fprintf(w, "<TABLE>\n<THEAD><TR><TH>Code</TH><TH>Description</TH></TR></THEAD>\n<TBODY>\n")
-	for _, p := range o.HttpResponses {
+	responses := o.HttpResponses
+	sort.Slice(responses, func(i, j int) bool {
+		return strings.Compare(responses[i].Name, responses[j].Name) < 0
+	})
```
kbhawkey: @tengqm , can you provide more information (or example output from the 1.17 version of the API ref) about the `fake changes` in the response section.

kbhawkey:
@tengqm , are you introducing a 1.16 release branch?

gen-apidocs/generators/html.go
tengqm: https://github.com/kubernetes/website/pull/18466/files#diff-f333bc8f279559f54f72b1ac02c65cecR7317

kbhawkey: I see that there are duplicates. Is this because there is a bug in the generation caused by a different order of the response code+description or some other reason (different copy of swagger.json)?

tengqm: I'm not sure, but if you are really curious, you can track the iteration here: https://github.com/kubernetes-sigs/reference-docs/blob/master/gen-apidocs/generators/api/config.go#L427

The `range` call is not returning items in fixed order.

kbhawkey: 👍

PR #128, opened 2020-01-10
Make sure API responses are sorted

tengqm:
xref: #126

### DIFF
--- a/gen-apidocs/generators/html.go
+++ b/gen-apidocs/generators/html.go
@@ -23,6 +23,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


kbhawkey:
See my comments from PR #126

tengqm:
@kbhawkey Yes we have a release-1.16 branch for regenerating the 1.16 references. This PR contains an improvement to master, which is also proposed to release-1.16. The idea behind this PR was triggered by a previous commit (https://github.com/kubernetes/website/pull/18466/files#diff-f333bc8f279559f54f72b1ac02c65cecR7317) where you can see many diffs are actually fake changes. The order of the response records was different.

kbhawkey:
@tengqm , I think you mean that duplication of content was added into the generated reference (not really faked by a human or added manually).

tengqm:
Yes. It is not a bug in the strict sense. But having responses yielded in different orders each time is not good.

kbhawkey:
> 
> 
> Yes. It is not a bug in the strict sense. But having responses yielded in different orders each time is not good.

OK, got it.
/lgtm

PR #130, opened 2020-01-17
clean up script and makefile

kbhawkey:
Removed write to tmp files and cat commands.
Retested script and build.


tengqm:
Nice. I was about to do this but failed to find a time slot.
/lgtm
/approve

PR #131, opened 2020-01-18
correct swagger.json fetch string

kbhawkey:
- correct git show fetch string
- fix variable names in script for config dir copying
- errors in fetch string not caught because swagger.json existed in version dir
- removed config/swagger.json, using 1.17.0 file


kbhawkey:
/cc @tengqm
If you have time to review.
I decided to double check the update api target. I realized that the target was succeeding (false positive) only because the file exists in the versioned directory. Also, the original (suggested) fetch/show string did not work.

Issue #118, opened 2020-01-07
migrate kubectl reference


kbhawkey:
This is an umbrella issue to capture ideas for updating the kubectl command reference.
- Related to PR #83.
- The kubectl reference is difficult to read and navigate and is not easily maintained.
- The kubectl reference generation is error prone and typically the generated output has encountered a number of text formatting issues.

Come up with a solution to stabilize the kubectl command reference generation, UI, and code maintenance. Coordinate changes with the API reference generation (share assets, similar UI, similar generation process). As an alternative, the kubectl command reference could generate as a group of pages (Markdown) similar to other k8s/website reference tool pages. Is it important to readers to maintain a `reference` for the kubectl commands?

tengqm:
One of the complaints we got is about the "examples". I'm not sure how useful they are but I'm unhappy about the fact that the examples layout is eating too much area of the screen.
For API reference, generating HTML directly is not optimal. There is still a tradeoff between having one huge HTML and having separate HTMLs for each group/kind/operation.
Generating markdown files from these generators and then generating HTMLs from these markdowns ... this doesn't make any senses to me. We may want to generate HTMLs (and other output formats) directly.
Consolidating all generators (API ref, kubectl ref, kube-* components...) to a single framework would be desirable.
We may need a redesign the framework so that:

- the same process can be used across all references;
- the look & feel of the generated (HTML) is consistent for all docs;
- the navigation across sections/subsections is made easier;
- the generation process is automated for users, especially the release team who may be not so familiar with the generators;
- the generation is highly customizable with configuration knobs exposed where appropriate.
- ...

kbhawkey:
I agree with you about directly generating HTML (not Markdown). Some more ideas:
- The kubectl examples could be added to the main page for the command instead of in a separate container.
- The references could use a single code base or part of a common code base -- same backend and similar look-and-feel. Improve the navigation.

kbhawkey:
I am temporarily assigning myself.
I started to update the code to directly generate HTML as well as clean up the css, js, toc.yaml. Will post what I have shortly.
/assign

tengqm:
xref: kubernetes/website#14904

tengqm:
xref: https://github.com/Birdrock/brodocs/pull/9 and the issue: https://github.com/kubernetes/website/issues/19519

PR #132, opened 2020-01-27
create simplified kubectl ref

kbhawkey:
- Built simple, single page kubectl command reference.
- [TODO] Description, Usage, Examples text. Really any descriptive text should be converted or cleaned up. I found that the text has a number of markup and special chars such as: important note (!!!), embedded links, embedded lists. Newlines could be preserved and whitespace. Possibly convert Markdown snippets to HTML (goldmark), then run through golang template.Html.
- [TODO] Overall page styling needs more work. Determine colors, highlight text, etc. Verify static assets and clean up.
- [TODO] Remove static-includes directory and files and update setup script.
- Flags are presorted (alphabetical); does this need additional verification.
- [TODO] Add table captions (Aria).
- Simplified Makefile
- See issue #118.


kbhawkey:
image

gen-kubectldocs/generators/html.go
```diff
@@ -0,0 +1,371 @@
...
+		for _, tlc := range g.Commands {
+			m[tlc.MainCommand.Name] = tlc
+		}
+	}
+
+	// REVISIT, use the toc.yaml file to map the tlc to the category/group
+	for _, c := range toc.Categories {
+
+		// Write each of the commands in this category
+		for _, cm := range c.Commands {
```
tengqm: any consideration on the ordering of values in the map?

tengqm: maybe it is okay at this stage ... eventually we need an ordered list of the commands and options so that the output can be easily compared to each other.

gen-kubectldocs/generators/html.go
```diff
@@ -0,0 +1,371 @@
...
+			"<th>Name</th>" +
+			"<th>Shorthand</th>" +
+			"<th>Default</th>" +
+			"<th>Usage</th>" +
+			"</tr>" +
+			"</thead>" +
+			"<tbody>"
+
+		for _, option := range cmd.Options {
+			buf += "<tr><td>" + option.Name + "</td><td>" + option.Shorthand + "</td><td>" +
```
tengqm: option.Name may need some tags like `<code>`.

gen-kubectldocs/generators/html.go
kbhawkey: Yes, thinking of sorting the commands based on the order/grouping of commands in toc.yaml. I updated toc.yaml to match the output of `kubectl help`.
Ideally, the group name/message would be provided by the spec (or kubectl help).

gen-kubectldocs/generators/html.go
kbhawkey: 👍

gen-kubectldocs/generators/html.go
kbhawkey: The commands are ordered per category. This is fixed in toc.yaml.
I guess the order of the sub-commands and options could change from release to release. The rate of change seems kind of slow, though.

tengqm: ok

tengqm:
@kbhawkey Is this still a WIP? Can we kick it in and iterate over it?

kbhawkey:
> 
> 
> @kbhawkey Is this still a WIP? Can we kick it in and iterate over it?

@tengqm , This is sort of a WIP. I started to work on cleaning up the Examples section, but have not gotten back to it yet. I was not sure if there was any interest in this PR.
I am definitely in favor a new process and UI for this reference.
If the examples were fixed, then I would consider this PR less of a WIP. There are a lot of style tweaks that could be made, but I am happy with the basic way the docs generate.
The description text, examples text, and description text of flags needs to be cleaned up (Markdown, odd formatting, etc.). I will reopen my k8s/website PR which shows the full reference built using the code from this PR. Have a look again and let me know.
Do you have other ideas?

tengqm:
@kbhawkey You are right we need to tune the look&feel of examples. There are other things to tune as well, such as the style for the left navigator. I was hoping that we can make it auto-hiding.
As a useful option, I'd vote for getting this in, even if it is not favored by the website team.

kbhawkey:
Since it has been some time, I will look at the state of the changes (and changes I made but did not merge in yet). 
Yes, I agree to iterate and move these changes along.

gen-kubectldocs/config/v1_17/toc.yaml
```diff
@@ -0,0 +1,61 @@
+categories:
+- name: GETTING STARTED
```
julianvmodesto: Around content organization, perhaps a grouping could be `APPLICATION MANAGEMENT`?

Going even further, I'd prefer to have groupings for `IMPERATIVE APPLICATION MANAGEMENT` and `DECLARATIVE APPLICATION MANAGEMENT` to mirror the corresponding doc sections under [Manage Kubernetes Objects](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/):

```
Declarative Management of Kubernetes Objects Using Configuration Files
Declarative Management of Kubernetes Objects Using Kustomize
Managing Kubernetes Objects Using Imperative Commands
Imperative Management of Kubernetes Objects Using Configuration Files
```

kbhawkey: @julianvmodesto, Should the group names be restricted to a specific length? The names are part of the left navigation.

Is it possible for `kubectl` to provide the groupings or categories. The toc could be dynamically generated instead of manually updated whenever the commands change.
I think there is an issue/PR in spf13/cobra requesting something like this (https://github.com/spf13/cobra/issues/836).

kbhawkey:
Before adding this in, is there time (v1.18 release) to get the reference to a point where it is no longer a draft and is an acceptable replacement for the current kubectl command reference (granted I think the current ref is difficult to use and looks outdated).

kbhawkey:
Placing a hold on this again.
/hold

PR #146, opened 2020-04-06
Fix example for type definitions

tengqm:
We are generating two buttons under the resource type definitions, for
'curl' and 'kubectl' respectively. However, for resource type
definitions, this logic is wrong. One examples for operations have
different types. This fix the problem.


kbhawkey:
@tengqm , Can you describe what this change will look like?

tengqm:
@kbhawkey Before change: 
Screen Shot 2020-04-07 at 9 51 11 AM
After change:
Screen Shot 2020-04-07 at 9 53 38 AM

kbhawkey:
Thanks @tengqm 
/lgtm

Issue #136, opened 2020-03-21
Tweak the page content style


tengqm:
The current 'page-content container' class assigned to the right (main) panel needs some  tweaking:

- The current style is wasting some real estate on screen.
- We'd better add some attributes to the tables so that the property name column is not too narrow (xref: kubernetes/website#19757)

Arhell:
@tengqm 
Hello, please see the solution
https://github.com/kubernetes/website/pull/19771
---------------------------------------------------------------------------------
https://deploy-preview-19771--kubernetes-io-master-staging.netlify.com/docs/reference/generated/kubernetes-api/v1.17/

Arhell:
ping @tengqm 
#137

tengqm:
Related: https://github.com/kubernetes/website/issues/19795

PR #147, opened 2020-04-07
Fix page grid

tengqm:
Closes: #136 

We are currently employing both custom sytlesheet and Bootstrap classes.
This is hurting the stability of the page layout. This PR proposes to
use the bootstrap grid system so that we don't need to customize the
20%/80% division manually. Hopefully this can yield a better layout that
works across different browsers.


kbhawkey:
@tengqm , I agree with "This PR proposes to
use the bootstrap grid system so that we don't need to customize"
I would like to see the rendered output.
/lgtm

Issue #141, opened 2020-03-25
verify config directory checks for kubectl generator


kbhawkey:
I ran into a problem when running the update-imported-docs script for the 1.18 release.
Previously, I set up the reference configuration directories because some of the references required configuration changes for the release (run make `createversiondirs`).
The gen-kubectldocs/generators/v1_18/toc.yaml was written over (typical behavior if release dir does not exist) instead of reusing the toc.yaml in the existing target directory.
Look into the logic in the setup script, set_version_dirs.sh.

Issue #148, opened 2020-04-08
Make HTML table output Commonmark compliant


kbhawkey:
In preparation for upgrading Hugo (Goldmark parser),
clean up the HTML table output for all component tool pages.
https://github.com/kubernetes-sigs/reference-docs/blob/master/gen-compdocs/generators/doc.go#L296

For additional information, see issue: https://github.com/kubernetes/website/pull/19907.

Issue #99, opened 2019-12-24
modify http-cache, kubectl, gen-compdocs


kbhawkey:
With Kubernetes v1.17, kubectl `cache-dir` default value contains a resolved value for $HOME.
Substitute resolved value with `$HOME`.
-cache-dir Default: "$HOME/.kube/http-cache"

tengqm:
Do you mean the `$HOME` one? I don't think we get a chance to see the special character. The default value was already set to current user's home directory before we know its existence.

kbhawkey:
Yes, `$HOME`. This issue is new with v1.17. Is it possible to change in k8s/k8s?

tengqm:
@kbhawkey I'm not optimistic about this. See the code here: 

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cli-runtime/pkg/genericclioptions/config_flags.go#L57

and where it is referenced before we check the config structure:

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cli-runtime/pkg/genericclioptions/config_flags.go#L326

tengqm:
xref: https://github.com/kubernetes/website/pull/18010#pullrequestreview-328878902

kbhawkey:
> @kbhawkey I'm not optimistic about this. See the code here:
> 
> https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cli-runtime/pkg/genericclioptions/config_flags.go#L57
> 
> and where it is referenced before we check the config structure:
> 
> https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cli-runtime/pkg/genericclioptions/config_flags.go#L326

OK. Thanks for clarifying.
Possibly clean up content in gen'ed table:

```
<tr>
      <td colspan="2">--cache-dir string&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Default: "/home/username/.kube/http-cache"</td>
    </tr>
    <tr>
      <td></td><td style="line-height: 130%; word-wrap: break-word;">Default HTTP cache directory</td>
    </tr>
```

PR #150, opened 2020-04-11
modify comp docs table indentation

kbhawkey:
- Fixed issue #141. The script incorrectly checked for `config.yaml` instead of `toc.yaml`.
- Added a solution for #99. This replaces the output of `$HOME` with the variable name "$HOME" for the value of the flag, `cache-dir`. Possibly moved into a post-processing function.
- Updated the `Options` HTML table block (`gen-compdocs/generators/doc.go`) indentation so that the inline HTML table renders. For information, see issue #148.
- The sample output can be found in pull request, https://github.com/kubernetes/website/pull/20245 .

### DIFF
--- a/gen-compdocs/generators/doc.go
+++ b/gen-compdocs/generators/doc.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/user"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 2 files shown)


tengqm:
Looks great. Thanks.
/lgtm
/approve

PR #160, opened 2020-06-20
Add "latest" version of API docs to provide more stable external link targets

joelsmith:
This is a companion PR to https://github.com/kubernetes/website/pull/21931
and provides a way to keep the "latest" API link target up to date.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,7 @@ K8SRELEASEDIR=$(shell echo "$(K8SRELEASE)" | sed "s/\./_/g")
 
 APISRC=gen-apidocs
 APIDST=$(WEBROOT)/static/docs/reference/generated/kubernetes-api/v$(K8SRELEASE)
+APILATEST=$(shell ( echo v$(K8SRELEASE); ls -d $(dir $(APIDST))/v[0-9]* ) | xargs basename -a | sort -V -r | head -1)
 
 CLISRC=gen-kubectldocs/generators/build
 CLIDST=$(WEBROOT)/static/docs/reference/generated/kubectl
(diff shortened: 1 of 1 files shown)


tengqm:
Great. This automates the changes to the redirects entry.
/lgtm
Defer approval to @kbhawkey

Makefile
```diff
@@ -16,6 +16,7 @@ K8SRELEASEDIR=$(shell echo "$(K8SRELEASE)" | sed "s/\./_/g")
 
 APISRC=gen-apidocs
 APIDST=$(WEBROOT)/static/docs/reference/generated/kubernetes-api/v$(K8SRELEASE)
+APILATEST=$(shell ( echo v$(K8SRELEASE); ls -d $(dir $(APIDST))/v[0-9]* ) | xargs basename -a | sort -V -r | head -1)
```
kbhawkey, line 19: Hi @joelsmith .
Just catching up and trying to put the pieces together. A couple of thoughts:
The gist of this change is to update the proposed `latest` entry in the website/static/_redirects file.
This target would be run after the docs are released?
I guess I have reservations about updating the _redirects file from a target in this Makefile.
I wonder if the docs can provide a more permanent solution by creating a `latest` page for you to 
reference? That way there is no need to modify the _redirects file.
I have not checked the targets.
APIDST should be set to something like:
`website/static/docs/reference/generated/kubernetes-api/v1.18`
I have not sorted out the details, but what do you think about creating a `latest` directory (or page)?

sftim:
Bear in mind that https://github.com/kubernetes/website/pull/21931 is held pending discussion

> I wonder if the docs can provide a more permanent solution by creating a latest page for you to reference? That way there is no need to modify the _redirects file.

seems like a nicer approach to me.

PR #165, opened 2020-07-30
Allowing generating API reference for RC versions

kbhawkey:
Testing option to build with k8s release candidate versions.
This example can generate the references. Still working on cleaning up rc_num string usage.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,9 @@
 # K8S_WEBROOT=~/src/github.com/kubernetes/website
 # K8S_ROOT=~/k8s/src/k8s.io/kubernetes
 # K8S_RELEASE=1.17
+# RC_NUM=-rc.3
 
+RCNUM=${RC_NUM}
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -6,7 +6,9 @@
 # K8S_WEBROOT=~/src/github.com/kubernetes/website
 # K8S_ROOT=~/k8s/src/k8s.io/kubernetes
 # K8S_RELEASE=1.17
+# RC_NUM=-rc.3
 
+RCNUM=${RC_NUM}
```
tengqm, line 11: instead of introducing additional vars, how about we relax the release number string to be '1.18.0', '1.18.1' etc?

kbhawkey: Yes, thinking on it.

kbhawkey: I'd like the option to use the release candidate tag before the release, otherwise
the root string is okay,

tengqm: OK

PR #169, opened 2020-07-31
clean up updateapispec target

kbhawkey:
Clean up open api spec copy target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,19 +5,20 @@
 #
 # K8S_WEBROOT=~/src/github.com/kubernetes/website
 # K8S_ROOT=~/k8s/src/k8s.io/kubernetes
-# K8S_RELEASE=1.17
-# RC_NUM=-rc.3
+# K8S_RELEASE=1.17.0, 1.17.5, 1.17.0-rc.2
+
 
 RCNUM=${RC_NUM}
 WEBROOT=${K8S_WEBROOT}
 K8SROOT=${K8S_ROOT}
 K8SRELEASE=${K8S_RELEASE}
+K8SRELEASE_PREFIX=$(shell echo "$(K8SRELEASE)" | cut --characters 1-4)
 
 # create a directory name from release string, e.g. 1.17 -> 1_17
-K8SRELEASEDIR=$(shell echo "$(K8SRELEASE)" | sed "s/\./_/g")
+K8SRELEASEDIR=$(shell echo "$(K8SRELEASE_PREFIX)" | sed "s/\./_/g")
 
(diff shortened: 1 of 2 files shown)


kbhawkey:
For some reason, the bot added an approved label (automatically).

kbhawkey:
Tested `updateapispec` target with k8s v1.18.0 and v1.19.0-rc.3.

tengqm:
While reviewing this change, I was thinking of our previous change adding RCNUM.
Maybe we should instead allow for two version string patterns rather than adding the RC number as a special case.

- 'x.yz': This is an abbreviated version of 'x.yz.0', and it is for convenience and backward compatibility;
- all other valid version tag strings, such a '1.18.6', '1.19.0-rc2' ... 

In the Makefile, we first check if the K8S_RELEASE is of format `x.yz` and append the `.0` suffix if so.
Or else we use the whole string as a version tag.

We then attempt to fetch the version tag from k8s repo and error if the tag is not found.

WDYT?

kbhawkey:
Re: In the Makefile, we first check if the K8S_RELEASE is of format x.yz and append the .0 suffix if so.
This makes a special case for the `.0` suffix. I am not opposed to using the full version of the release tag for all cases.
However, the existing directory structures (config, build dirs in this repo and static dir in website) limit the way the different versions are handled and stored. In other words, v1.19-rc.3 is built as copied to `website/static` as v1.19 not v1.19-rc.3 or v1.19.5.
I think this needs more thought. but is heading in the right direction to handle building patched and release candidate versions.
How about adjusting the directory structure after the 1.19 release?

kbhawkey:
@tengqm . I am investigating using any format (1.xy.z, 1.xy.z-rc.x) for the K8S_RELEASE value.
There are minor changes to getting the directory name to match the existing structure for building/copying.
The config and website/static directory structures could remain as is (for now). For example `config/v1_18`.
Is this what you had in mind? The update-imported* script would use the actual tag name to clone the k8s/k8s repo
and update the open api spec. I'd like to avoid using 1.17 and instead use the fully specified version name such as 1.17.0.

kbhawkey:
I need to test the latest changes. The update* script would remain the same. I'd like to change `reference.yml` to
accept `$(K8S_RELEASE)` in full instead of `$ (K8S_RELEASE).0`.

tengqm:
This looks a lot cleaner than the previous version. Let me know when the test is completed.

kbhawkey:
I tested these changes with PR #22898. The API config directory and `website/static` directories remain the same.

PR #172, opened 2020-10-09
Add support for generating API group version list

tengqm:
This has been a desired feature on kubernetes/website. (kubernetes/website#2979)

### DIFF
--- a/gen-apidocs/generators/html.go
+++ b/gen-apidocs/generators/html.go
@@ -79,6 +79,49 @@ func (h *HTMLWriter) WriteOverview() {
 	h.CurrentSection = &item
 }
 
+func (h *HTMLWriter) WriteAPIGroupVersions(gvs api.GroupVersions) {
+	fn := "_group_versions.html"
+	path := filepath.Join(api.IncludesDir, fn)
+	f, err := os.Create(path)
+	defer f.Close()
+	if err != nil {
+		os.Stderr.WriteString(fmt.Sprintf("%v", err))
+		os.Exit(1)
+	}
+
+	fmt.Fprintf(f, h.DefaultStaticContent("API Groups"))
+	fmt.Fprintf(f, "<P>The API Groups and their versions are summarized in the following table.</P>")
+	fmt.Fprintf(f, "<TABLE class=\"col-md-8\">\n<THEAD><TR><TH>Group</TH><TH>Version</TH></TR></THEAD>\n<TBODY>\n")
+
+	groups := api.ApiGroups{}
+	for group, _ := range gvs {
+		groups = append(groups, api.ApiGroup(group))
+	}
+	sort.Sort(groups)
+
+	for _, group := range groups {
+		versionList, _ := gvs[group.String()]
+		sort.Sort(versionList)
+		var versions []string
+		for _, v := range versionList {
+			versions = append(versions, v.String())
+		}
+
+		fmt.Fprintf(f, "<TR><TD><CODE>%s</CODE></TD><TD><CODE>%s</CODE></TD></TR>\n",
+			group, strings.Join(versions, ", "))
+	}
+	fmt.Fprintf(f, "</TBODY>\n</TABLE>\n")
+
+	item := TOCItem{
+		Level: 1,
+		Title: "API Groups",
+		Link:  "-strong-api-groups-strong-",
+		File:  fn,
+	}
+	h.TOC.Sections = append(h.TOC.Sections, &item)
+	h.CurrentSection = &item
+}
+
 func (h *HTMLWriter) WriteResourceCategory(name, file string) {
 	writeStaticFile(name, "_"+file+".html", h.DefaultStaticContent(name))
 	link := strings.Replace(strings.ToLower(name), " ", "-", -1)
--- a/gen-apidocs/generators/api/definition.go
+++ b/gen-apidocs/generators/api/definition.go
@@ -31,11 +31,11 @@ import (
 // inlineDefinition is a definition that should be inlined when displaying a Concept
 // instead of appearing the in "Definitions"
 type inlineDefinition struct {
-	Name string
+	Name  string
 	Match string
 }
 
-var _INLINE_DEFINITIONS = []inlineDefinition {
+var _INLINE_DEFINITIONS = []inlineDefinition{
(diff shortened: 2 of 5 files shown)


tengqm:
image

gen-apidocs/generators/html.go
```diff
@@ -79,6 +79,49 @@ func (h *HTMLWriter) WriteOverview() {
...
+	f, err := os.Create(path)
+	defer f.Close()
+	if err != nil {
+		os.Stderr.WriteString(fmt.Sprintf("%v", err))
+		os.Exit(1)
+	}
+
+	fmt.Fprintf(f, h.DefaultStaticContent("API Groups"))
+	fmt.Fprintf(f, "<P>The API Groups and their versions are summarized in the following table.</P>")
+	fmt.Fprintf(f, "<TABLE class=\"col-md-8\">\n<THEAD><TR><TH>Group</TH><TH>Version</TH></TR></THEAD>\n<TBODY>\n")
```
irvifa, line 94: Can we show what kind of format we have so that we need to print this formatting? Or probably write the expected output as comment on the function.

tengqm: See my comment above. It is a screenshot

sftim:
Likely to be relevant to https://github.com/kubernetes/website/issues/23809

gen-apidocs/generators/html.go
```diff
@@ -79,6 +79,49 @@ func (h *HTMLWriter) WriteOverview() {
 	h.CurrentSection = &item
 }
 
+func (h *HTMLWriter) WriteAPIGroupVersions(gvs api.GroupVersions) {
+	fn := "_group_versions.html"
+	path := filepath.Join(api.IncludesDir, fn)
+	f, err := os.Create(path)
+	defer f.Close()
+	if err != nil {
```
kbhawkey, line 87: Hi @tengqm . Nice to see different ways of presenting the information. Wondering about:
The table appears after the overview on the page?
Approximately how many unique groups (rows) are in the table? There are no links in the table and the table does not scroll or collapse.
Would it be useful to list or link to the resources found under each group (Not in the scope of this PR but may be something to think about)?

tengqm: There are about 20 rows in the table. It is a very simple table. Listing resources under each group is doable.

kbhawkey: Hi @tengqm . Since this is information is requested by readers, it makes sense to provide and highlight the groups
and supported versions.
Suggestion: Could this information be presented as a standalone single page or could the information be inferred
by updating the API navigation. Does it make sense to navigate by group (not category)?
I think the new information could be created, pushed/copied to the site repo, included in a md page or use a writer to create a suppored docs page layout with this content.
What do you think? The changes could go in with follow-up changes. 
Let's temporarily place this on hold (feel free to cancel the hold after discussion).
/hold

tengqm: Surely we can dump this data into whatever format once we have the map.
I chose to summarize this API group/version info when we are extracting definitions from the swagger spec so we don't need to traverse it again.
Since the only source of truth is the swagger spec, I'm not sure what do you mean by "could the information be inferred ...". 
My immediate response (could be wrong) to your question "navigating by group" is a No. API groups (and versions) are never the first thing a user wants to check. This was true at least in the past when we have `Deployment` defined in both `extensions` group and `apps` group, `Ingress` defined in both `extensions` and `networking` group, etc. As the upstream community is gradually hardening the "core" resource types to support, the grouping of resources are much more clearer. Navigating resources by groups might be eventually something to consider.
Another fact we need to consider is that not all swagger definitions are resources, though they are also very important to users. If we put API groups on the top level, we will have to list `apimachinery`, `meta`, ... there as group names. I'm pretty sure it will drive some users crazy.
Back to your last question, splitting this information into a separate file is doable. The only concern I have is about maintenance overhead. We are already blamed by someone for the inferior quality of the generator. We have to be cautious when adding new files to manage for each release.
With that said, if we can settle down on a solution that generates 100 markdown files and copy them in batch to website, this information can be a standalone page shipped together with others.

kbhawkey: > Surely we can dump this data into whatever format once we have the map.
> I chose to summarize this API group/version info when we are extracting definitions from the swagger spec so we don't need to traverse it again.

-- OK

> Since the only source of truth is the swagger spec, I'm not sure what do you mean by "could the information be inferred ...".

-- I wonder if the table is needed if the existing information is presented differently.


> My immediate response (could be wrong) to your question "navigating by group" is a No. API groups (and versions) are never the first thing a user wants to check. This was true at least in the past when we have `Deployment` defined in both `extensions` group and `apps` group, `Ingress` defined in both `extensions` and `networking` group, etc. As the upstream community is gradually hardening the "core" resource types to support, the grouping of resources are much more clearer. Navigating resources by groups might be eventually something to consider.

--OK. I just thought it would make sense since "groups" are built in categories.

> Another fact we need to consider is that not all swagger definitions are resources, though they are also very important to users. If we put API groups on the top level, we will have to list `apimachinery`, `meta`, ... there as group names. I'm pretty sure it will drive some users crazy.

-- Could filter out some groups?


> Back to your last question, splitting this information into a separate file is doable. The only concern I have is about maintenance overhead. We are already blamed by someone for the inferior quality of the generator. We have to be cautious when adding new files to manage for each release.
> With that said, if we can settle down on a solution that generates 100 markdown files and copy them in batch to website, this information can be a standalone page shipped together with others.

-- I am thinking about Hugo modules (https://gohugo.io/hugo-modules/) instead of copying files.

tengqm: The reason I'm summarizing the API groups and versions table is that people need such information for specifying the `--runtime-config` flag (https://github.com/kubernetes/website/issues/2979). This data *can be inferred* from existing page by taking note whenever you see a new API group version. In fact, that is exactly what this PR is doing. We can automatically summarize the supported API group versions into one table. If you don't know what API group version can be enabled/disabled, just check this table.
When we have this table in place, you can reference it from any place in the website.

I'm not familiar with Hugo modules.

kbhawkey: OK to add this now, though, when the big page is separated, this table could move to another page.
/lgtm
/hold cancel

PR #179, opened 2020-12-02
New resourcesdocs generator

feloy:
This generator is used to generate API Reference pages in Markdown format.

See https://github.com/kubernetes/website/pull/23294 for more details


gen-resourcesdocs/Makefile
```diff
@@ -0,0 +1,9 @@
+clean:
+	rm -rf kwebsite/content/en/docs/* kwebsite/public
+
+kwebsite: clean
+	mkdir -p kwebsite/content/en/docs
+	go run cmd/main.go kwebsite --config-dir config/v1.20/ --file api/v1.20/swagger.json --output-dir kwebsite/content/en/docs --templates ./templates
```
sftim, line 6: Would it make sense to take the version as a parameter somehow?

sftim (commented):
I'm -0 (soft opposed) to including `gen-resourcesdocs/api/v1.19/swagger.json` and `gen-resourcesdocs/api/v1.20/swagger.json`; but that's something we could clean up later.

I'm also wary of including the configuration inside `gen-resourcesdocs/config/v1.20/` in the same commit as the code that consumes it. These are two related but separate changes.
LGTM, but @feloy if you were willing to split this into multiple commits then I think that works even better.

I'd like someone familiar with Golang to confirm this change is OK to merge.

feloy:
> I'm -0 (soft opposed) to including `gen-resourcesdocs/api/v1.19/swagger.json` and `gen-resourcesdocs/api/v1.20/swagger.json`; but that's something we could clean up later.
> 

It would be nice to make the build reproducible by being sure which swagger file has been used. Do you prefer we put the swagger file on the k/website repo wih the generated files?

kbhawkey:
It would be really nice if there was a job that created a temporary k8s/k8s repo, grabbed a version of the spec, copied the spec to this repo or website.
Some of the code to create this job already exists in update-imported-docs script+reference.yml (create k8s/k8s repo) which calls `updateapispec` target (fetch swagger.json) in reference-docs/Makefile.

tengqm:
/approve

I'm inclined to kick this in now and start the iteration.

kbhawkey:
I think it makes sense to get this in. I read through the code, though I have not tried to build the generator.
LGTM!!

kbhawkey:
@sftim, What do you think about getting the source code in now? The swagger files can be removed when the build targets are updated. Is the plan to publish the 1.20 pages generated from the source?

sftim:
@kbhawkey I agree with your LGTM from https://github.com/kubernetes-sigs/reference-docs/pull/179#issuecomment-737627010

With the caveat that **I** _haven't_ done code review and _haven't_ much Golang experience.
/lgtm

sftim:
@feloy many congratulations :tada::tada::tada:

feloy:
> @feloy many congratulations 🎉🎉🎉

Thanks to everybody 🎆🎆🎆

kbhawkey:
Super!

zacharysarah:
@feloy Congratulations! ✨ 🌟 ✨

sftim:
I logged https://github.com/kubernetes/website/issues/25505 to track switching [k/website](https://github.com/kubernetes/website) to use the new style API reference.

Issue #178, opened 2020-12-02
Update API reference generator for 1.20 release


kbhawkey:
Output from running:
```
cd $K8S_WEBSITE/update-imported-docs
 ./update-imported-docs.py reference.yml 1.20.0-rc.0
```
```
No Definition found for readInternalApiserverV1alpha1StorageVersion [/apis/internal.apiserver.k8s.io/v1alpha1/storageversions/{name}].  
No Definition found for deleteInternalApiserverV1alpha1StorageVersion [/apis/internal.apiserver.k8s.io/v1alpha1/storageversions/{name}].  
No Definition found for patchInternalApiserverV1alpha1StorageVersion [/apis/internal.apiserver.k8s.io/v1alpha1/storageversions/{name}].  
No Definition found for replaceInternalApiserverV1alpha1StorageVersion [/apis/internal.apiserver.k8s.io/v1alpha1/storageversions/{name}].  
No Definition found for deleteInternalApiserverV1alpha1CollectionStorageVersion [/apis/internal.apiserver.k8s.io/v1alpha1/storageversions].  
No Definition found for createInternalApiserverV1alpha1StorageVersion [/apis/internal.apiserver.k8s.io/v1alpha1/storageversions].  
No Definition found for listInternalApiserverV1alpha1StorageVersion [/apis/internal.apiserver.k8s.io/v1alpha1/storageversions].  
No Definition found for watchInternalApiserverV1alpha1StorageVersionList [/apis/internal.apiserver.k8s.io/v1alpha1/watch/storageversions].  
No Definition found for getServiceAccountIssuerOpenIDKeyset [/openid/v1/jwks/].  
No Definition found for patchInternalApiserverV1alpha1StorageVersionStatus [/apis/internal.apiserver.k8s.io/v1alpha1/storageversions/{name}/status].  
No Definition found for replaceInternalApiserverV1alpha1StorageVersionStatus [/apis/internal.apiserver.k8s.io/v1alpha1/storageversions/{name}/status].  
No Definition found for readInternalApiserverV1alpha1StorageVersionStatus [/apis/internal.apiserver.k8s.io/v1alpha1/storageversions/{name}/status].  
No Definition found for getServiceAccountIssuerOpenIDConfiguration [/.well-known/openid-configuration/].  
No Definition found for watchInternalApiserverV1alpha1StorageVersion [/apis/internal.apiserver.k8s.io/v1alpha1/watch/storageversions/{name}].  
----------------------------------
[...]

PR #180, opened 2020-12-02
updates to build 1.20 references

kbhawkey:
- Removed cloud-controller-manager from build target
- Created configuration for v1.19 and v1.20
- Update kubectl command toc.yaml (removed convert, added debug)


sftim (commented):
As this is for v1.20 @kbhawkey can you confirm that files such as `gen-kubectldocs/generators/v1_19/static_includes/_app_management.md` supposed to be part of this change?

sftim:
Maybe the v1.19 change could be a separate commit to make that more obvious?

kbhawkey:
Yes, I ran the Makefile twice (1.19 and 1.20) to generate the configuration files (target createversiondirs).
The script relies upon a previous configuration to generate the next configuration and does not default to any configuration file.
The 1.19 build used the 1.18 configuration (generates dynamically in the build sandbox but does not commit).
I never got around to generating and merging a 1.19 configuration. I did this now.
I have not read through the API configuration. The API reference will need code/config updates.

kbhawkey:
I have not looked at updating the API generator/config to address issue #178.

kbhawkey:
There is one more error to resolve for `v1alpha1/storageversions`.
/cc @tengqm

tengqm:
/lgtm
/approve

> There is one more error to resolve for v1alpha1/storageversions.

I'll check it.

tengqm:
@kbhawkey Fix is proposed in #181

PR #181, opened 2020-12-03
Fix API generation for v1.20 (rc.0)

tengqm:
This PR fixes two problems in the API reference generator:

1. The `StorageVersion` resource introduced in 1.20 is not a good citizen:

- resource API version named: `internal.apiserver.k8s.io`
- definition path in Swagger: `io.k8s.api.apiserverinternal`
- operation ID defined as: `createInternalApiserver...`

Note that the two key words ("internal" and "apiserver") are in different orders; sometimes they split, other times they are merged.

2. The following APIs are not RESTful APIs because they are not associated with any REST resources. These two APIs have to be skipped.

- `getServiceAccountIssuerOpenIDConfiguration` exposed at `/.well-known/openid-configuration/`;
- `getServiceAccountIssuerOpenIDKeyset` exposed at `/openid/v1/jwks`.

closes: #178

### DIFF
--- a/gen-apidocs/config/v1_20/config.yaml
+++ b/gen-apidocs/config/v1_20/config.yaml
@@ -2,6 +2,7 @@ example_location: "examples"
 api_groups:
   - "AdmissionRegistration"
   - "ApiExtensions"
+  - "InternalApiserver"
   - "ApiRegistration"
   - "Apps"
   - "AuditRegistration"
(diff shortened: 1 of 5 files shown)


kbhawkey:
Changes look good.
What about logging an issue in `kubernetes/kubernetes`?
/lgtm
/approve

PR #183, opened 2020-12-03
Expose group mapping into config file.

tengqm:
The group names used in operation IDs are sometimes different from the group names seen by the corresponding resource. This PR expose such mappings for customization when bad citizens show up in the API spec.

/hold

Holding until #181 is in.

### DIFF
--- a/gen-apidocs/config/v1_20/config.yaml
+++ b/gen-apidocs/config/v1_20/config.yaml
@@ -321,3 +321,9 @@ group_full_names:
   scheduling: scheduling.k8s.io
   storage: storage.k8s.io
 
+# The map from the group as the resource sees it to the group as the operation
+# sees it.
+operation_group_map:
+  rbac: RbacAuthorization
+  flowcontrol: FlowcontrolApiserver
+  apiserverinternal: InternalApiserver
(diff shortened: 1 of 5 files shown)


gen-apidocs/config/v1_20/config.yaml
```diff
@@ -271,3 +276,11 @@ operation_categories:
       match: create${group}${version}(Namespaced)?${resource}Rollback
     - name: Read Log
       match: read${group}${version}(Namespaced)?${resource}Log
+
+# The map from the group as the resource sees it to the group as the operation
+# sees it.
+operation_group_map:
+  rbac: RbacAuthorization
+  flowcontrol: FlowcontrolApiserver
```
kbhawkey, line 284: OK. This is configurable instead of programmatically "munging" the group?

gen-apidocs/generators/api/operation.go
```diff
@@ -134,8 +136,10 @@ func (o *Operation) VerifyBlackListed() {
 	case strings.Contains(o.ID, "V1beta1CertificateSigningRequestApproval"):
 	case strings.Contains(o.ID, "V1CertificateSigningRequestApproval"):
 	case strings.Contains(o.ID, "V1beta1NamespacedReplicationControllerDummyScale"):
+	case strings.Contains(o.ID, "getServiceAccountIssuerOpenIDConfiguration"):
```
kbhawkey, line 139: These strings will be removed and replaced by configuration?

tengqm:
/hold cancel
Dependent PR merged. Conflicts resolved.

PR #184, opened 2020-12-03
Allow customization of the full names for groups

tengqm:
Groups have different schemes for naming its full group names. This is bad but we have to live with it. This PR avoids hardcoding this conversion or guessing by exposing the full names for users to adapt.

### DIFF
--- a/gen-apidocs/generators/api/definition.go
+++ b/gen-apidocs/generators/api/definition.go
@@ -45,14 +45,14 @@ var _INLINE_DEFINITIONS = []inlineDefinition{
 	{Name: "EventSource", Match: "${resource}EventSource"},
 }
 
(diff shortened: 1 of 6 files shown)


tengqm:
/hold
until #181 is merged.

gen-apidocs/generators/api/definition.go
```diff
@@ -222,12 +222,17 @@ func (d *Definition) GroupDisplayName() string {
 }
 
 func (d *Definition) GetOperationGroupName() string {
+	// TODO(Qiming): Expose this to config.yaml so that we don't need to
+	// change the Go source next time.
```
kbhawkey, line 226: OK. This gets cleaned up.

tengqm:
#181 is in
/hold cancel

PR #186, opened 2020-12-11
resourcedocs: Add Anchors to inlined definitions

feloy:
Inlined definitions into the Markdown generated by gen-resourcesdocs are not referenceable.

These changes adds anchors to the inlined definitions.

These changes also explicitly set fragment names for Resources.

### DIFF
--- a/gen-resourcesdocs/pkg/outputs/interface.go
+++ b/gen-resourcesdocs/pkg/outputs/interface.go
@@ -28,7 +28,7 @@ type Chapter interface {
 // Section is an interface to a section of an output
 type Section interface {
 	AddContent(s string) error
-	AddTypeDefinition(s string) error
+	AddTypeDefinition(typ string, description string) error
 	StartPropertyList() error
 	AddFieldCategory(name string) error
 	AddProperty(name string, property *kubernetes.Property, linkend []string, indent bool, defname string, shortName string) error
--- a/gen-resourcesdocs/pkg/outputs/kwebsite/files.go
+++ b/gen-resourcesdocs/pkg/outputs/kwebsite/files.go
@@ -63,7 +63,6 @@ func escapeName(parts ...string) string {
 
 // headingID returns the ID built by hugo for a given header
 func headingID(s string) string {
-	result := strings.ToLower(s)
-	result = strings.ReplaceAll(result, " ", "-")
+	result := strings.ReplaceAll(s, " ", "-")
 	return result
 }
--- a/gen-resourcesdocs/templates/chapter-single-definition.tmpl
+++ b/gen-resourcesdocs/templates/chapter-single-definition.tmpl
@@ -22,6 +22,7 @@ weight: {{.Metadata.Weight}}
 {{.Description | replace "<" "\\<" | indent 2 | indent .Indent | indent .Indent}}
 {{- end}}
 {{if .TypeDefinition}}
+{{ ""  | indent .Indent | indent .Indent}}  <a name="{{.Type}}"></a>
 {{.TypeDefinition | indent 2 | indent .Indent | indent .Indent}}
 {{end}}
 {{- end}}{{/* range .Fields */}}
@@ -35,6 +36,7 @@ weight: {{.Metadata.Weight}}
 {{.Description | replace "<" "\\<" | indent 2 | indent .Indent | indent .Indent}}
 {{- end}}
 {{if .TypeDefinition}}
+{{ ""  | indent .Indent | indent .Indent}}  <a name="{{.Type}}"></a>
 {{.TypeDefinition | indent 2 | indent .Indent | indent .Indent}}
 {{end}}
 {{- end}}{{/* range .Fields */}}
--- a/gen-resourcesdocs/templates/chapter-single-definition.tmpl
+++ b/gen-resourcesdocs/templates/chapter-single-definition.tmpl
@@ -28,7 +28,7 @@ weight: {{.Metadata.Weight}}
 {{- end}}{{/* range .Fields */}}
 
 {{range .FieldCategories}}
-### {{.Name}}
+### {{.Name}} {#{{.Name}}}
 
 {{range .Fields}}
 {{ ""  | indent .Indent | indent .Indent}}- {{.Name}}{{if .Value}}: {{.Value}}{{end}}
--- a/gen-resourcesdocs/templates/chapter-single-definition.tmpl
+++ b/gen-resourcesdocs/templates/chapter-single-definition.tmpl
@@ -28,7 +28,7 @@ weight: {{.Metadata.Weight}}
 {{- end}}{{/* range .Fields */}}
 
 {{range .FieldCategories}}
-### {{.Name}} {#{{.Name}}}
+### {{.Name}} {#{{.Name}}}{{/* explicitly set fragment to keep capitalization */}}
 
 {{range .Fields}}
 {{ ""  | indent .Indent | indent .Indent}}- {{.Name}}{{if .Value}}: {{.Value}}{{end}}
--- a/gen-resourcesdocs/pkg/config/output.go
+++ b/gen-resourcesdocs/pkg/config/output.go
@@ -212,7 +212,7 @@ func (o *TOC) OutputProperties(defname string, definition spec.Schema, outputSec
 						return err
 					}
 
-					err = outputSection.AddTypeDefinition(target.Description)
+					err = outputSection.AddTypeDefinition(property.TypeKey.ResourceName(), target.Description)
 					if err != nil {
 						return err
 					}
--- a/gen-resourcesdocs/pkg/config/output_test.go
+++ b/gen-resourcesdocs/pkg/config/output_test.go
@@ -32,9 +32,9 @@ func (o FakeChapter) Write() error { return nil }
 
 type FakeSection struct{}
 
-func (o FakeSection) AddContent(s string) error          { return nil }
-func (o FakeSection) AddTypeDefinition(s string) error   { return nil }
-func (o FakeSection) AddFieldCategory(name string) error { return nil }
+func (o FakeSection) AddContent(s string) error                              { return nil }
+func (o FakeSection) AddTypeDefinition(typ string, description string) error { return nil }
+func (o FakeSection) AddFieldCategory(name string) error                     { return nil }
 
 func (o FakeSection) AddProperty(name string, property *kubernetes.Property, linkend []string, indent bool, defname string, shortName string) error {
 	return nil
--- a/gen-resourcesdocs/pkg/outputs/kwebsite/chapter.go
+++ b/gen-resourcesdocs/pkg/outputs/kwebsite/chapter.go
@@ -51,6 +51,7 @@ type FieldData struct {
 	Name           string
 	Value          string
 	Description    string
+	Type           string
 	TypeDefinition string
 	Indent         int
 }
--- a/gen-resourcesdocs/pkg/outputs/kwebsite/section.go
+++ b/gen-resourcesdocs/pkg/outputs/kwebsite/section.go
@@ -24,7 +24,7 @@ func (o Section) AddContent(s string) error {
 }
 
 // AddTypeDefinition adds the definition of a type to a section
-func (o Section) AddTypeDefinition(s string) error {
+func (o Section) AddTypeDefinition(typ string, description string) error {
 	i := len(o.chapter.data.Sections)
 	cats := o.chapter.data.Sections[i-1].FieldCategories
 	var fields *[]FieldData
@@ -34,7 +34,8 @@ func (o Section) AddTypeDefinition(s string) error {
 		fields = &cats[len(cats)-1].Fields
 	}
 	j := len(*fields)
-	(*fields)[j-1].TypeDefinition = "*" + s + "*"
+	(*fields)[j-1].Type = typ
+	(*fields)[j-1].TypeDefinition = "*" + description + "*"
 	return nil
 }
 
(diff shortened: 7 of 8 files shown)


gen-resourcesdocs/templates/chapter-single-definition.tmpl
```diff
@@ -22,19 +22,21 @@ weight: {{.Metadata.Weight}}
...
 {{- end}}
 {{if .TypeDefinition}}
+{{ ""  | indent .Indent | indent .Indent}}  <a name="{{.Type}}"></a>
 {{.TypeDefinition | indent 2 | indent .Indent | indent .Indent}}
 {{end}}
 {{- end}}{{/* range .Fields */}}
 
 {{range .FieldCategories}}
-### {{.Name}}
+### {{.Name}} {#{{.Name}}}
```
sftim, line 31: Maybe add a comment about why the fragment identifier is set?

sftim (commented):
It passed CI, and the intention makes sense to me.
/lgtm

kbhawkey:
👀

gen-resourcesdocs/pkg/outputs/interface.go
```diff
@@ -28,7 +28,7 @@ type Chapter interface {
 // Section is an interface to a section of an output
 type Section interface {
 	AddContent(s string) error
-	AddTypeDefinition(s string) error
+	AddTypeDefinition(typ string, description string) error
```
kbhawkey, line 31: nit: `typ` is the type string?

gen-resourcesdocs/pkg/outputs/kwebsite/files.go
```diff
@@ -63,7 +63,6 @@ func escapeName(parts ...string) string {
 
 // headingID returns the ID built by hugo for a given header
 func headingID(s string) string {
-	result := strings.ToLower(s)
-	result = strings.ReplaceAll(result, " ", "-")
+	result := strings.ReplaceAll(s, " ", "-")
```
kbhawkey, line 66: Is it necessary to normalize the heading?
I was under the impression that Hugo would do this.

gen-resourcesdocs/pkg/outputs/interface.go
feloy: Yes, as `type` is a reserved keyword in Go, we cannot use `type` as variable name

gen-resourcesdocs/pkg/outputs/kwebsite/files.go
feloy: Yes, Hugo does it by default, but puts the anchor in lowercase. I would prefer to keep the original case, as these are Resource names, and I would like to use the anchor name as text to display in the api-reference shortcode: (see https://github.com/kubernetes/website/pull/23294/files#diff-4992971bcbda91c2f2374a2e62fffda3d7852174780e027070e766c62bf87ac1 )

gen-resourcesdocs/templates/chapter-single-definition.tmpl
kbhawkey: Do you need the set the fragment identifier if the same as the heading?
Hugo should automatically generate the anchors for the headings.

feloy: See my comment below,I would like to keep the original case of the heading (which is a Resource Name)

gen-resourcesdocs/pkg/outputs/interface.go
kbhawkey: Right, `typ` is odd. `t` or something else might make it more readable.

gen-resourcesdocs/pkg/outputs/kwebsite/files.go
kbhawkey: I need to look at the generated pages.
Could you update the shortcode to adapt the formatting of the resource name?

gen-resourcesdocs/pkg/outputs/interface.go
feloy: I think this is quite common in Go, see for example https://golang.org/pkg/go/types/

kbhawkey: Ha, interesting.

gen-resourcesdocs/templates/chapter-single-definition.tmpl
```diff
@@ -22,19 +22,21 @@ weight: {{.Metadata.Weight}}
...
 {{- end}}
 {{if .TypeDefinition}}
+{{ ""  | indent .Indent | indent .Indent}}  <a name="{{.Type}}"></a>
 {{.TypeDefinition | indent 2 | indent .Indent | indent .Indent}}
 {{end}}
 {{- end}}{{/* range .Fields */}}
 
 {{range .FieldCategories}}
-### {{.Name}}
+### {{.Name}} {#{{.Name}}}{{/* explicitly set fragment to keep capitalization */}}
```
sftim, line 31: :+1: to the comment, thanks @feloy

gen-resourcesdocs/pkg/outputs/kwebsite/files.go
feloy: The shortcode has been changed in this commit: https://github.com/kubernetes/website/pull/23294/commits/d9b18d378815988960f503e47ec8c0a8d32b3feb

The result should be visible in the paragraph https://deploy-preview-23294--kubernetes-io-master-staging.netlify.app/docs/concepts/configuration/secret/#using-imagepullsecrets  (see the PodSpec API ...)

The build of the preview website fails, I can't see why

kbhawkey:
What happens when this code builds in CI? Are there files generated into a staging directory in this repository?
For some time now, I have wanted a way to test tooling changes by building the website with the generated output.

gen-resourcesdocs/pkg/outputs/kwebsite/files.go
kbhawkey: looking now. See this message:
This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository.

kbhawkey: You need to rebase?

feloy: I think I made a rebase to restart the tests in the meantime. Here is the commit:
https://github.com/kubernetes/website/pull/23294/commits/99e1b4122221066583700d710228070988d75234

feloy: I also get the message, but you can see the change in the api-reference.html file of this commit

kbhawkey:
Looking at:
https://deploy-preview-23294--kubernetes-io-master-staging.netlify.app/docs/reference/kubernetes-api/workloads-resources/pod-v1/#lifecycle

And:
```
"io.k8s.api.core.v1.Lifecycle": {
      "description": "Lifecycle describes actions that the management system should take in response to container lifecycle events. For the PostStart and PreStop lifecycle handlers, management of the container blocks until the action is complete, unless the container process fails, in which case the handler is aborted.",
      "properties": {
        "postStart": {
          "$ref": "#/definitions/io.k8s.api.core.v1.Handler",
          "description": "PostStart is called immediately after a container is created. If the handler fails, the container is terminated and restarted according to its restart policy. Other management of the container blocks until the hook completes. More info: https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#container-hooks"
        },
        "preStop": {
          "$ref": "#/definitions/io.k8s.api.core.v1.Handler",
          "description": "PreStop is called immediately before a container is terminated due to an API request or management event such as liveness/startup probe failure, preemption, resource contention, etc. The handler is not called if the container crashes or exits. The reason for termination is passed to the handler. The Pod's termination grace period countdown begins before the PreStop hooked is executed. Regardless of the outcome of the handler, the container will eventually terminate within the Pod's termination grace period. Other management of the container blocks until the hook completes or until the termination grace period is reached. More info: https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#container-hooks"
        }
```

Why is ,
```
 "terminationGracePeriodSeconds": {
          "description": "Optional duration in seconds the pod needs to terminate gracefully. May be decreased in delete request. Value must be non-negative integer. The value zero indicates delete immediately. If this value is nil, the default grace period will be used instead. The grace period is the duration in seconds after the processes running in the pod are sent a termination signal and the time when the processes are forcibly halted with a kill signal. Set this value longer than the expected cleanup time for your process. Defaults to 30 seconds.",
          "format": "int64",
          "type": "integer"
        },
```

Found under the heading `Lifecycle`?
How are the fields ordered on the page for `PodSpec`?

feloy:
> Looking at:
> https://deploy-preview-23294--kubernetes-io-master-staging.netlify.app/docs/reference/kubernetes-api/workloads-resources/pod-v1/#lifecycle
> 
> [...]
> 
> Found under the heading `Lifecycle`?
> How are the fields ordered on the page for `PodSpec`?

Hi @kbhawkey the headings are defined in the configuration file at (1)

Note that the heading Lifecyle is different from the `Lifecycle` Kubernetes resource.

feloy:
> What happens when this code builds in CI? Are there files generated into a staging directory in this repository?
> It would be great to see the generated Markdown files.
> For some time now, I have wanted a way to test tooling changes by building the website with the generated output.

The changes are not visible in this repo, but you can see the changes in the PR (1), especially in the commits (2) and (3)


(1) https://github.com/kubernetes/website/pull/23294
(2) https://github.com/kubernetes/website/pull/23294/commits/7c4fc1c4d0ff4835033285b96701b0b730f47396
(3) https://github.com/kubernetes/website/pull/23294/commits/dffa26485a72ad46622ab53fc18581e504b3026d

kbhawkey:
> 
> 
> > Looking at:
> > https://deploy-preview-23294--kubernetes-io-master-staging.netlify.app/docs/reference/kubernetes-api/workloads-resources/pod-v1/#lifecycle
> > [...]
> > Found under the heading `Lifecycle`?
> > How are the fields ordered on the page for `PodSpec`?
> 
> Hi @kbhawkey the headings are defined in the configuration file at (1)
> 
> Note that the heading Lifecyle is different from the `Lifecycle` Kubernetes resource.
> 
> (1) https://github.com/kubernetes-sigs/reference-docs/blob/master/gen-resourcesdocs/config/v1.20/fields.yaml#L22

Hi @feloy 
OK, this is what I thought. I found it confusing to use a Kubernetes object name as a heading if the fields under that heading do not represent the object's fields. Compare this representation to fields such as `affinity`, which refer/resolve to the `Affinity` object.
With the `Affinity` field, the sub-fields are in-lined below the field name but the reader does not get a hint that `Affinity` is an actual object (could use a link as well as inlining the object definition?).
See `"$ref": "#/definitions/io.k8s.api.core.v1.Affinity",`. Is there a common definition for `Affinity`? I see that there are common definitions for `NodeAffinity`, `PodAffinity`, and so on.

How did you select the text for the sub-headings? Would it help to list the fields alphabetically?
What are the headings `Alpha level`, `Beta level`, `Deprecated`? Does this mean that the fields under this heading are not GA?

I am reviewing this page:
https://deploy-preview-23294--kubernetes-io-master-staging.netlify.app/docs/reference/kubernetes-api/workloads-resources/pod-v1/#PodSpec

* It think it would help provide a link for every field (are HTML headings used?). If the fields used headings (instead of a list), then Hugo would add anchors. The generator could add anchors relative to the main headings on the page.
* I noticed that some links in the text are not normalized. See the readinessGates field in `PodSpec`,
   `https://git.k8s.io/enhancements/keps/sig-network/0007-pod-ready%2B%2B.md`

kbhawkey:
Should I log an issue (change template) for separating the resource description (that contains a list) from the list of fields?
Possibly add a separator between the top description and the body of the resource (fields). Otherwise, the two lists blend together.

For example, `StatefulSet`:

```
StatefulSet represents a set of pods with consistent identities. Identities are defined as:

    * Network: A single stable DNS and hostname.

    * Storage: As many VolumeClaims as requested. The StatefulSet guarantees that a given network identity will always map to the same storage identity.
```
https://deploy-preview-23294--kubernetes-io-master-staging.netlify.app/docs/reference/kubernetes-api/workloads-resources/stateful-set-v1/#StatefulSet

feloy:
Yes, please create an issue for this problem, I didn't see this problem before.

kbhawkey:
Some more thoughts about intra-page links:

It would be useful to link to any field in an object or resource from another page.
Right now, the fields are list items and a heading serves as the link to the set of fields contained under a heading.
These headings are not directly related to other k8s objects/resources but instead represent configurable categories per page.

What do you think about creating links for each field/list item?
The categories help divide the page and add navigable content, but it is not clear that the names of the categories help in understanding the object hierarchy and information about the fields.

kbhawkey:
> 
> 
> Yes, please create an issue for this problem, I didn't see this problem before.

👍 See issue #188.

feloy:
> Some more thoughts about intra-page links:
> 
> It would be useful to link to any field in an object or resource from another page.
> Right now, the fields are list items and a heading serves as the link to the set of fields contained under a heading.
> These headings are not directly related to other k8s objects/resources but instead represent configurable categories per page.
> 
> What do you think about creating links for each field/list item?
> The categories help divide the page and add navigable content, but it is not clear that the names of the categories help in understanding the object hierarchy and information about the fields.

Yes, I agree.
When I updated the links in the Concepts and Tasks sections to the new API reference, I realized lots of links were targeting specific fields. For the moment, I made the links target the category to which the field belongs, but it would be nice to have targetable fields.

feloy:
I'm not sure, is there something to do on my side for this PR?

Issue #193, opened 2021-01-05
migrate site build from travis


kbhawkey:
- Investigate how to replace travis with prow or G. actions. Seeking a lightweight solution.
- Build the individual reference doc generator tool source code per pull request.
- Match the release version with latest release version in kubernetes/website/config.toml.
- Create a script or reuse the setup script to pull the release information.

PR #195, opened 2021-01-05
[wip]update travis script

kbhawkey:
Updated `.travis.yml` script to use Go 1.15.5, Kubernetes 1.20.
Updated library dependencies and commands to build:
- component tool pages
- kubectl command reference
- gen-resourcedocs

**Note**: Need to enable commands (library dependencies) to build current API reference.
Related to issue #193.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ install:
 script:
 - make comp
 - make genresources
-- make updateapispec
+#- make updateapispec
 - make cli
 
 # TODO: uncomment the following to make other docs
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,29 @@
 env:
   global:
-    - KUBE_VERSION: 1.16
+    - KUBE_VERSION: 1.20
+    - K8S_RELEASE: 1.20
+
 language: go
 go:
-  - 1.12.1
+  - 1.15.5
 
 install:
 - export PATH=$GOPATH/bin:$PATH
 - mkdir -p $GOPATH/src/k8s.io/kubernetes
 - wget https://dl.k8s.io/v${KUBE_VERSION}.0/kubernetes-src.tar.gz -P $GOPATH/src/k8s.io/kubernetes
 - pushd $GOPATH/src/k8s.io/kubernetes && tar xzf kubernetes-src.tar.gz && popd
-- pushd $GOPATH/src/k8s.io/kubernetes && make generated_files && cp -L -R vendor ${GOPATH}/src/ && rm -r vendor && popd
+- pushd $GOPATH/src/k8s.io/kubernetes && make generated_files && cp -L -R vendor ${GOPATH}/src/ && popd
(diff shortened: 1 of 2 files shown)


kbhawkey:
Hi @zacharysarah .
Do you if it is possible to swap out travis for another build tool?

zacharysarah:
@kbhawkey 
> Do you if it is possible to swap out travis for another build tool?

It should be possible, yes. I can't think of any dependencies that require Travis specifically. 

Are you thinking of a specific replacement yet, or just curious in general?

kbhawkey:
I am looking into updating the build tool (prow).

kbhawkey:
I updated the travis script to build the component tool pages, kubectl command reference, gen-resourcesdocs.
The script depends on the correct release version.
I need to add more commands to build the `api` target. Ideally, each reference would build in a separate container/job
as there are library conflicts (note some libraries are removed from vendor directory). 
The build takes a long time (currently just under 10 minutes?)!

My intention is to move this build/pull request job to prow. Other recommendations welcome.

tengqm:
Thanks for this change, @kbhawkey 
I also spent some time yesterday debugging the build issue. The root cause is that GoLang is transiting from the old GOPATH based dependency to GO MOD based dependency. We have been hacking the k/k repo for its vendor directories.
I'm wondering if we are eventually going to use go mod as well.

.travis.yml
```diff
@@ -1,20 +1,29 @@
...
 
 install:
 - export PATH=$GOPATH/bin:$PATH
 - mkdir -p $GOPATH/src/k8s.io/kubernetes
 - wget https://dl.k8s.io/v${KUBE_VERSION}.0/kubernetes-src.tar.gz -P $GOPATH/src/k8s.io/kubernetes
 - pushd $GOPATH/src/k8s.io/kubernetes && tar xzf kubernetes-src.tar.gz && popd
-- pushd $GOPATH/src/k8s.io/kubernetes && make generated_files && cp -L -R vendor ${GOPATH}/src/ && rm -r vendor && popd
+- pushd $GOPATH/src/k8s.io/kubernetes && make generated_files && cp -L -R vendor ${GOPATH}/src/ && popd
+- pushd $GOPATH/src/k8s.io/kubernetes && rm -r vendor/github.com/spf13/cobra && rm -r vendor/github.com/spf13/pflag && popd
+- pushd $GOPATH/src/k8s.io/kubernetes && cp -R vendor/k8s.io/kubectl/pkg/cmd  pkg/kubectl/ && popd
```
tengqm, line 17: no. this hack is too dirty.
We may want to change the import from k8s.io/kubectl/pkg/cmd directly, rather than from k8s.io/kubernetes/pkg/kubectl/cmd.

kbhawkey: Yes, agreed.
The source for the reference doc generators should be built and tested before merging.
The library problems/vendor hacking was a problem with the last several releases. It is an issue.
I think it is worthwhile to change/update the build tool and improve the build/test processes.

kbhawkey:
Closing as PR #197 updated the component source and Makefile targets.

PR #197, opened 2021-01-07
revise gencomp and travis

tengqm:


kbhawkey:
@tengqm .
The changes improve the build of the component pages.
I still think it is a good idea to change CI/CD tools (these changes can be moved over).
What do you think about adding in the changes from (#195 ) to build `gen-resourcesdocs`?
Should be straightforward.

- Add `gen-resourcesdocs` target to .travis.yml
- Update the Makefile with a target such as:
```
genresources:
	cd gen-resourcesdocs && $(MAKE) kwebsite
```

gen-compdocs/go.mod
```diff
@@ -0,0 +1,54 @@
...
+	github.com/yuin/goldmark-highlighting v0.0.0-20200307114337-60d527fdb691 // indirect
+	golang.org/x/mod v0.3.0 // indirect
+	golang.org/x/tools v0.0.0-20200616133436-c1934b75d054 // indirect
+	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
+	gopkg.in/yaml.v2 v2.3.0 // indirect
+	k8s.io/gengo v0.0.0-20201113003025-83324d819ded // indirect
+	k8s.io/klog/v2 v2.4.0 // indirect
+	k8s.io/kubectl v0.0.0
+	k8s.io/kubernetes v0.0.0
+	// k8s.io/kubernetes v1.20.0
```
kbhawkey, line 24: ?

tengqm: That line should be removed.

tengqm:
@kbhawkey Adding the missing pieces.

kbhawkey:
@tengqm , The changes look good.
Do the targets run only if there are changes to be tested? I assumed each target would be evaluated as part of the merge check (regardless of changes to the source).

```
make gen-resourcesdocs

make: Nothing to be done for 'gen-resourcesdocs'.

The command "make gen-resourcesdocs" exited with 0.
```

tengqm:
@karenb  According to your suggestion, the top-level make target was `genresources`, not `gen-resourcesdocs`.

Currently, we are running the targets we specified in the .travis.yml. These targets are irrelevant to the changes made in a specific PR.

kbhawkey:
> @karenb According to your suggestion, the top-level make target was `genresources`, not `gen-resourcesdocs`.
> 

Hi @tengqm .

-- I proposed that you add a target (could name whatever you want) for buliding `gen-resourcesdocs` from the main project Makefile.


> Currently, we are running the targets we specified in the .travis.yml. These targets are irrelevant to the changes made in a specific PR.

-- Yes. I proposed that you add the correct target in .travis.yml to build `gen-resourcesdocs`. I was asking if the travis build was set to build if changes were detected or to build always.

tengqm:
@kbhawkey DO you mean this? https://github.com/kubernetes-sigs/reference-docs/blob/master/Makefile#L89-L90

kbhawkey:
Yes, I _think_ we are referring to the same thing (not sure).

tengqm:
@kbhawkey So the genresources target is working and is included in the travis config.

Makefile
```diff
@@ -11,7 +11,7 @@
 RCNUM=${RC_NUM}
 WEBROOT=${K8S_WEBROOT}
 K8SROOT=${K8S_ROOT}
-K8SRELEASE=${K8S_RELEASE}
+K8SRELEASE=${K8S_RELEASE:1.20.0}
```
benley, line 14: Does this substitution syntax actually work in a Makefile context?  It appears that it's trying to expand an undefined variable if I run make with `--warn-undefined-variables`:

```
Makefile:35: warning: undefined variable 'K8S_RELEASE:1.20.0'
```

kbhawkey: The original variable substitution worked correctly (set `K8S_RELEASE`).
I have not tested this syntax:`{K8S_RELEASE:1.20.0}`.
cc @tengqm

tengqm: looks like my fault, will confirm and fix if needed.

PR #200, opened 2021-01-13
add build.yaml, workflow

kbhawkey:
Testing a workflow to build reference documentation and replace `.travis.yml`.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -18,6 +18,9 @@ jobs:
 
     runs-on: ${{ matrix.platform }}
 
+    env:
+      BUILD_PATH: ${{ github.workspace }}/tmp
+
     steps:
 
       - name: Set up Go ${{ matrix.go-version }}
@@ -32,24 +35,20 @@ jobs:
       - name: Set release version
         id: version_value
         run: |
-          mkdir -p ${BUILD_PATH}
-          cd ${BUILD_PATH} && wget https://raw.githubusercontent.com/kubernetes/website/master/config.toml
+          mkdir -p ${{env.BUILD_PATH}}
+          cd ${{env.BUILD_PATH}} && wget https://raw.githubusercontent.com/kubernetes/website/master/config.toml
           echo 'KUBE_VERSION<<EOF' >> $GITHUB_ENV
-          awk '/^version/ {print}' ${BUILD_PATH}/config.toml | cut -c 12-16 >> $GITHUB_ENV
+          awk '/^version/ {print}' ${{env.BUILD_PATH}}/config.toml | cut -c 12-16 >> $GITHUB_ENV
           echo 'EOF' >> $GITHUB_ENV
-        env:
-          BUILD_PATH: ${{ github.workspace }}/tmp
 
       - name: Build references
         run: |
           echo "Building comp and gen-resourcesdocs"
           echo "$GITHUB_WORKSPACE"
-          mkdir -p ${BUILD_PATH}/src/k8s.io
-          echo "${BUILD_PATH}"
-          wget https://github.com/kubernetes/kubernetes/archive/${KUBE_VERSION}.0.tar.gz -O ${BUILD_PATH}/src/k8s.io/kubernetes-src.tar.gz
-          pushd ${BUILD_PATH}/src/k8s.io && tar xzf kubernetes-src.tar.gz && mv kubernetes-${KUBE_VERSION}.0 kubernetes && popd
-          pushd ${BUILD_PATH}/src/k8s.io/kubernetes && make generated_files && popd
+          mkdir -p ${{env.BUILD_PATH}}/src/k8s.io
+          echo "${{env.BUILD_PATH}}"
+          wget https://github.com/kubernetes/kubernetes/archive/${{env.KUBE_VERSION}}.0.tar.gz -O ${{env.BUILD_PATH}}/src/k8s.io/kubernetes-src.tar.gz
+          pushd ${{env.BUILD_PATH}}/src/k8s.io && tar xzf kubernetes-src.tar.gz && mv kubernetes-${{env.KUBE_VERSION}}.0 kubernetes && popd
+          pushd ${{env.BUILD_PATH}}/src/k8s.io/kubernetes && make generated_files && popd
           make comp
           make genresources
-        env:
-          BUILD_PATH: ${{ github.workspace }}/tmp
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,51 @@
+name: build-refdocs
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+  pull_request:
+
+jobs:
+  build:
+    name: Build reference docs
+
+    strategy:
+      matrix:
+        go-version: [1.15.x]
+        platform: [ubuntu-latest]
+
+    runs-on: ${{ matrix.platform }}
+
+    steps:
+
+      - name: Set up Go ${{ matrix.go-version }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -0,0 +1,34 @@
...
+      BUILD_PATH: /tmp
+    steps:
+      - name: Set up Go ${{ matrix.go-version }}
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go-version }}
+        id: go
+      - name: Build
+        run: |
+          mkdir -p ${BUILD_PATH}/src/k8s.io
```
tengqm, line 29: Do we have BUILD_PATH set somewhere?
In travis, we have global variables. Is it possible to do that in this workflow?

kbhawkey: I used /tmp. I am not sure what to do here.
I would like to run a script to read a configuration file containing say, `version`?

kbhawkey: Could add a config.toml containing BUILD_PATH and RELEASE_VERSION?

tengqm: It would be good to have configuration separated from the code (yaml workflow), although I'm not sure how to do that with workflows.

kbhawkey: Yes. Nice to read in a property file.

config.toml
```diff
@@ -0,0 +1,5 @@
+# configuration for automated builds
+
+latest = "v1.20"
+fullversion = "v1.20.0"
```
irvifa, line 4: When will we use the `fullversion` vs the `version`

config.toml
```diff
@@ -0,0 +1,5 @@
+# configuration for automated builds
```
zacharysarah, line 1: This is good to separate out from the YAML, per above. ☝🏻 

This configuration requires manual updates during every release cycle. If we add the update to the release cycle playbook, does that also mean configuring repo permissions for the release lead and their team to make and approve changes?

.github/workflows/build.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Build references
+        run: |
+          echo "Building comp and gen-resourcesdocs"
+          echo "$GITHUB_WORKSPACE"
+          mkdir -p ${BUILD_PATH}/src/k8s.io
+          echo "${BUILD_PATH}"
+          wget https://github.com/kubernetes/kubernetes/archive/${KUBE_VERSION}.0.tar.gz -O ${BUILD_PATH}/src/k8s.io/kubernetes-src.tar.gz
+          pushd ${BUILD_PATH}/src/k8s.io && tar xzf kubernetes-src.tar.gz && mv kubernetes-${KUBE_VERSION}.0 kubernetes && popd
+          pushd ${BUILD_PATH}/src/k8s.io/kubernetes && make generated_files && popd
```
irvifa, line 47: I'm not sure about this but when we want to move to specific directory and only for that line we can use:
```
(cd ${BUILD_PATH}/src/k8s.io/kubernetes && make generated_files)
```

config.toml
kbhawkey: I am not sure if it is needed.

config.toml
kbhawkey: Ideally, I'd like to use the configuration from `website` and not maintain a separate copy here.
The "action" would read (or pull in) config.toml.

.github/workflows/build.yaml
kbhawkey: I will look at the installation commands again.
Have I used the correct syntax to read and set environment variables in "actions" workflows?

config.toml
kbhawkey: Oh. The full version could be used in the installation string. Right now a '0' is appended.

.github/workflows/build.yaml
irvifa: I think we should use `env.` eg `${{ env.BUILD_PATH} }`

- https://github.community/t/how-to-set-and-access-a-workflow-variable/17335
- https://docs.github.com/en/actions/reference/environment-variables
This should be inline with the one for [secret](https://github.com/kubernetes/website/pull/25771/files#diff-0f9a227f719fb077ec8062bccb5a9db75597db3a96d6a159be72cb52f4e506c7R13)

kbhawkey: OK. Thanks for the information.
I looked at: https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable.
I can set BUILD_PATH in `jobs.build.env` assuming this variable has access to GITHUB_WORKSPACE. Alternatively, could create the installation at the workspace root and not use BUILD_PATH.
I set KUBE_VERSION in a step assuming that any later steps have access to this environment value as $KUBE_VERSION.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+      - name: Check out reference-docs code
+        uses: actions/checkout@v2
+
+      - name: Set release version
+        id: version_value
+        run: |
+          mkdir -p ${{env.BUILD_PATH}}
+          cd ${{env.BUILD_PATH}} && wget https://raw.githubusercontent.com/kubernetes/website/master/config.toml
+          echo 'KUBE_VERSION<<EOF' >> $GITHUB_ENV
```
sftim, line 40: should this be:
```suggestion
          echo 'KUBE_VERSION=<<EOF' >> $GITHUB_ENV
```
?

.github/workflows/build.yaml
```diff
@@ -0,0 +1,54 @@
...
+        run: |
+          echo "Building comp and gen-resourcesdocs"
+          echo "$GITHUB_WORKSPACE"
+          mkdir -p ${{env.BUILD_PATH}}/src/k8s.io
+          echo "${{env.BUILD_PATH}}"
+          wget https://github.com/kubernetes/kubernetes/archive/${{env.KUBE_VERSION}}.0.tar.gz -O ${{env.BUILD_PATH}}/src/k8s.io/kubernetes-src.tar.gz
+          pushd ${{env.BUILD_PATH}}/src/k8s.io && tar xzf kubernetes-src.tar.gz && mv kubernetes-${{env.KUBE_VERSION}}.0 kubernetes && popd
+          pushd ${{env.BUILD_PATH}}/src/k8s.io/kubernetes && make generated_files && popd
+          make comp
+          make genresources
```
sftim, line 54: After the build, what should happen? Is the aim of the Action to test that a build works / do we want to save the output?

kbhawkey: Right now, travis captures the build output and displays success or failure? 
I don't know if there is similar functionality available with GitHub Actions.

.github/workflows/build.yaml
kbhawkey: I based this change on these examples:
https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#multiline-strings
https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#example-6

.github/workflows/build.yaml
sftim: We'd need a place to put the output artifact(s) I think. I don't believe GitHub offers that bit.

kbhawkey: I think it is okay to build the source and report success/failure (do I need to add a code coverage action).
You could imagine saving the generated Markdown files (as artifacts) and adding a new step to run `hugo serve`.
My main focus at this point is to swap out the .travis script and build the source files.
Not sure if the G.A. builds will be any faster.

kbhawkey:
Closing as there is low interest in switching out Travis.
/close

Issue #198, opened 2021-01-13
Add comment to all generated files


tengqm:
Now and then we receive direct change requests to the generated files.
We should add a comment to all files we generate to indicate that "This is a auto-generated file, please do not change it directly."

kbhawkey:
I think it is helpful to create an informational string in the page footer ("auto-generated by <generator>").
I'm not sure I would go as far as asking not to pursue changes to the page.
If the Api reference resource pages use a custom layout, then we could prevent the display of the link
to the default "edit the page/create PR for page" template? This should be straightforward to change.

tengqm:
Masking out the 'edit/create' buttons would be good.
I'm talking about adding some comments to the generated files in source, so that people will know that this source file is not intended to be manually edited.
For example, https://github.com/kubernetes/website/pull/26035, https://github.com/kubernetes/website/pull/26069,

kbhawkey:
+1

jihoon-seo:
It seems that this issue is resolved by #201, and okay to close.
Is this right? :)

tengqm:
Right. This one can be closed.
/close

jihoon-seo:
Update:
Files like [content/en/docs/reference/command-line-tools-reference/kube-proxy.md](https://raw.githubusercontent.com/kubernetes/website/master/content/en/docs/reference/command-line-tools-reference/kube-proxy.md) has the comment
(`The file is auto-generated ...`),

but these files don't, even if these files are generated by the `update-imported-docs.py` script too.

- (Links below are from the PR https://github.com/kubernetes/website/pull/27189, which reflects the updated generated files.)
- [content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm.md](https://raw.githubusercontent.com/kubernetes/website/4ecb418a3baa362ceceff7ff73d0607bc23489fa/content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm.md)
- [content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_alpha.md](https://raw.githubusercontent.com/kubernetes/website/4ecb418a3baa362ceceff7ff73d0607bc23489fa/content/en/docs/reference/setup-tools/kubeadm/generated/kubeadm_alpha.md)
- ...

---

@tengqm Could you please take a look again?
Or is this (= 'Don't add the comment to `content/en/docs/reference/setup-tools/kubeadm/generated/*`') an intended behavior of #201?
Sorry for causing confusion!

PR #201, opened 2021-01-14
Add warnings in the generated files

tengqm:
Related: #198

### DIFF
--- a/gen-compdocs/generators/doc.go
+++ b/gen-compdocs/generators/doc.go
@@ -36,6 +36,18 @@ func (s byName) Len() int           { return len(s) }
 func (s byName) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
 func (s byName) Less(i, j int) bool { return s[i].Name() < s[j].Name() }
 
+const generated_warning = `<!--
+The file is auto-generated from the Go source code of the component using a generic
+[generator](https://github.com/kubernetes-sigs/reference-docs/). To learn how
+to generate the reference documentation, please read
+[Contributing to the reference documentation](/docs/contribute/generate-ref-docs/).
+To update the reference conent, please follow the 
+[Contributing upstream](/docs/contribute/generate-ref-docs/contribute-upstream/)
+guide. You can file document formatting bugs against the
+[reference-docs](https://github.com/kubernetes-sigs/reference-docs/) project.
+-->
+`
+
 func GenMarkdownTree(cmd *cobra.Command, dir string, withTitle bool) error {
 	identity := func(s string) string { return s }
 	emptyStr := func(s string) string { return "" }
@@ -85,9 +97,12 @@ func GenMarkdownCustom(cmd *cobra.Command, w io.Writer, linkHandler func(string)
 	// Note: Files generated for kubeadm tool are snippets of Markdown without a title.
 	// These snippets are included in the corresponding kubeadm pages.
 	if withTitle {
-		if _, err := fmt.Fprintf(w, "---\ntitle: %s\ncontent_type: tool-reference\nweight: 30\n---\n\n", name); err != nil {
+		if _, err := fmt.Fprintf(w, "---\ntitle: %s\ncontent_type: tool-reference\nweight: 30\nauto_generated: true\n---\n\n", name); err != nil {
 			return err
 		}
+
+		// Print the "generated" warning
+		fmt.Fprintf(w, "%s\n\n", generated_warning)
 
 		if _, err := fmt.Fprintf(w, "%s\n\n", "## {{% heading \"synopsis\" %}}"); err != nil {
 			return err


gen-compdocs/generators/doc.go
```diff
@@ -36,6 +36,17 @@ func (s byName) Len() int           { return len(s) }
 func (s byName) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
 func (s byName) Less(i, j int) bool { return s[i].Name() < s[j].Name() }
 
+const generated_warning = `<!--
+The file is auto-generated from the component source code. Please check
+[the overall introduction](/docs/contribute/generate-ref-docs/) for generating
+reference docs and the
```
kbhawkey, line 42: Could this source share between all reference tools?

gen-compdocs/generators/doc.go
```diff
@@ -85,10 +96,13 @@ func GenMarkdownCustom(cmd *cobra.Command, w io.Writer, linkHandler func(string)
 	// Note: Files generated for kubeadm tool are snippets of Markdown without a title.
 	// These snippets are included in the corresponding kubeadm pages.
 	if withTitle {
-		if _, err := fmt.Fprintf(w, "---\ntitle: %s\ncontent_type: tool-reference\nweight: 30\n---\n\n", name); err != nil {
+		if _, err := fmt.Fprintf(w, "---\ntitle: %s\ncontent_type: tool-reference\nweight: 30\ndoc_type: generated\n---\n\n", name); err != nil {
 			return err
 		}
 
+		// Print the "generated" warning
+		fmt.Fprintf(w, "%s\n\n", generated_warning)
```
kbhawkey, line 104: This warning appears immediately after the title as a comment.
nit: Instead of introducing `doc_type`, what do you think about adding a boolean? Something like: `auto-generated: true`?

gen-compdocs/generators/doc.go
```diff
@@ -36,6 +36,17 @@ func (s byName) Len() int           { return len(s) }
 func (s byName) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
 func (s byName) Less(i, j int) bool { return s[i].Name() < s[j].Name() }
 
+const generated_warning = `<!--
+The file is auto-generated from the component source code. Please check
```
kbhawkey, line 40: Hi @tengqm . Here is some feedback. Reword as you like.

This file is auto-generated from the [component reference source code](https://github.com/kubernetes-sigs/reference-docs/gen-compdocs).  To learn how to generate the reference documentation, read [Contributing to the reference documentation](/docs/contribute/generate-ref-docs/). To update the reference content, follow the [Contribute upstream](/docs/contribute/generate-ref-docs/contribute-upstream/) guide. You can file document formatting bugs against the [reference-docs](https://github.com/kubernetes-sigs/reference-docs/) project.

gen-compdocs/generators/doc.go
tengqm: That might be possible. Can revisit this when revising other tools.

gen-compdocs/generators/doc.go
tengqm: Yes. I'm afraid putting the warning at the very head of the page may disrupt Docsy.
Boolean tag is a good idea.
Once we have this tag, we can introduce some intelligence (as you suggested before) into Hugo so that the generated page doesn't have those 'edit' buttons.

gen-compdocs/generators/doc.go
tengqm: Will take your suggestion.

gen-compdocs/generators/doc.go
```diff
@@ -85,10 +97,13 @@ func GenMarkdownCustom(cmd *cobra.Command, w io.Writer, linkHandler func(string)
 	// Note: Files generated for kubeadm tool are snippets of Markdown without a title.
 	// These snippets are included in the corresponding kubeadm pages.
 	if withTitle {
-		if _, err := fmt.Fprintf(w, "---\ntitle: %s\ncontent_type: tool-reference\nweight: 30\n---\n\n", name); err != nil {
+		if _, err := fmt.Fprintf(w, "---\ntitle: %s\ncontent_type: tool-reference\nweight: 30\nauto-generated: true\n---\n\n", name); err != nil {
```
kbhawkey, line 100: This should be okay, though, would check if there is a convention for writing data fields as:
`auto_generated` or `autoGenerated` instead of `auto-generated`

kbhawkey: Hi @tengqm . I would fix the new field name, otherwise, the changes look good.
/approve
/hold

tengqm: let me put it as auto_generated.

kbhawkey: :+1:

PR #205, opened 2021-01-26
Fix defaulting in Makefile

tengqm:
Follow up of #197

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,12 @@
 RCNUM=${RC_NUM}
 WEBROOT=${K8S_WEBROOT}
 K8SROOT=${K8S_ROOT}
-K8SRELEASE=${K8S_RELEASE:1.20.0}
+K8SRELEASE=${K8S_RELEASE}
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -11,7 +11,12 @@
 RCNUM=${RC_NUM}
 WEBROOT=${K8S_WEBROOT}
 K8SROOT=${K8S_ROOT}
-K8SRELEASE=${K8S_RELEASE:1.20.0}
+K8SRELEASE=${K8S_RELEASE}
+
+ifeq ($(origin K8SRELEASE), undefined)
+K8SRELEASE=1.20.0
```
kbhawkey, line 17: nit: Is there a reason to pin to this version number?

tengqm: Well I was just trying to give it a default value if not assigned, or else the makefile won't work.
We can instead test and exit if required envvars are empty.
Any suggestions?

kbhawkey: I see. Sure, let's get a fix in now. Would be better to exit if the variables are not set.

PR #209, opened 2021-02-12
Use consistent pluralization for API groupings in reference

feloy:
Fixes https://github.com/kubernetes/website/issues/26361

The result can be seen at https://github.com/kubernetes/website/pull/26481


kbhawkey:
Thanks @feloy .
/lgtm

PR #176, opened 2020-11-24
Add reference generation tool to the repo

tengqm:
This tool can be used to generate *unpublished* "API" types. Check the README.md for more details.

PR #210, opened 2021-02-18
Add outputs from genref tool

tengqm:
I'm proposing a version control for the config outputs because:

- We want to know how frequent the upstream config structs are changed;
- The config files (HTML plus MD) are not huge assets to manage;
- We can later decide to remove this from version control when we gain
  more confidence about the generator and the workflow.

These files are separated from the PR #176 to ease the PR review.

### DIFF
--- a/genref/config.yaml
+++ b/genref/config.yaml
@@ -38,6 +38,8 @@ apis:
     title: Kubelet Stats (v1alpha1)
     package: k8s.io/kubelet
     path: pkg/apis/stats/v1alpha1
+    # Skip this because not referenced anywhere and it has no group name
+    skip: true
 
   - name: kube-scheduler-config
     title: kube-scheduler Configuration (v1beta1)
@@ -69,6 +71,8 @@ apis:
     title: Kubernetes Metrics (v1alpha1)
     package: k8s.io/metrics
     path: pkg/apis/metrics/v1alpha1
+    # superceded by v1beta1 version
+    skip: true
 
   - name: metrics
     title: Kubernetes Metrics (v1alpha1)
@@ -79,6 +83,8 @@ apis:
     title: Client Authentication (v1alpha1)
     package: k8s.io/client-go
     path: pkg/apis/clientauthentication/v1alpha1
+    # superceded by v1beta1
+    skip: true
 
   - name: client-authentication
     title: Client Authentication (v1beta1)
--- a/genref/markdown/members.tpl
+++ b/genref/markdown/members.tpl
@@ -4,21 +4,21 @@
   {{- range .GetMembers -}}
     {{/* . is a apiMember */}}
     {{- if not .Hidden }}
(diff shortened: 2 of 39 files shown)


tengqm:
/hold

this PR depends on #176.

kbhawkey:
👋 @tengqm .
I'd like more information about storing the generated files in this repository.
Will you tag or reference a specific commit after the files are generated (at release time)?
Will you replace the files every release or when the upstream files change?

tengqm:
> Will you tag or reference a specific commit after the files are generated (at release time)?
The outputs are supposed to be regenerated at each release, in a similar way how we have been handling kube- components.

> Will you replace the files every release or when the upstream files change?
The current plan is to replace the files every release.

> Will you save both Markdown and HTML versions of the files?
For markdown version, it is designed to be fed to the website. The HTML version is for local preview. I don't have a good solution to preview markdown locally as web pages.

> How many files are currently published?
Currently, there are 7 files proposed to k/website. We may need to maintain about 10 config API files eventually.

> Will you add the build of the tool to the base Makefile in this repository (and add to travis)?
Yes. That is the plan.

kbhawkey:
Hi. I added the docsy theme to this repository. It should help to evaluate newly rendered pages.
Let me know if the sub module is not set up correctly.
If you run `git submodule update --init --recursive --depth 1`, the theme should install.
You could copy the generated files to the reference section/add to the content tree, and test?

tengqm:
@kbhawkey Good idea. Will try.

kbhawkey:
Should be docsy-example. There is a `References` section.

tengqm:
The submodule command doesn't work for me. Tried several ways to work around it but failed.

```
docsy-example $ git submodule update --init --recursive --depth 1
fatal: No url found for submodule path 'themes/docsy' in .gitmodules
```

kbhawkey:
> 
> 
> The submodule command doesn't work for me. Tried several ways to work around it but failed.
> 
> ```
> docsy-example $ git submodule update --init --recursive --depth 1
> fatal: No url found for submodule path 'themes/docsy' in .gitmodules
> ```

OK. This is not working as expected. The "docsy-example" is not a sub-module. The theme is.
I was trying to add a "test" hugo site into this repository.
Newly generated reference pages could be copied into this content tree for testing.
I'll look into what's there now and see if I can sort it out.

tengqm:
/hold cancel
The generator is now in.

tengqm:
/hold
Holding this again because we may want to get #214, #215 in first.

tengqm:
/hold cancel
Removing hold because #176 is in.

tengqm:
/hold
Hold again because this is now depending on #215

kbhawkey:
These changes look ready?
/lgtm
Remove hold if ready.
Thanks @tengqm 
/hold

PR #213, opened 2021-02-23
Tune flag type for component references

tengqm:
The current generator exposes GoLang specific data types for command line arguments, e.g. `stringSlice`, `stringArray`, `mapStringString`. These type name come from the pflags package and they fail to accurately document the actual type of arguments for users to specify from command line.

This PR fixes this problem by replacing these data type names with a descriptive text where appropriate so that users have a better idea what are expected by each flag.

### DIFF
--- a/gen-compdocs/generators/doc.go
+++ b/gen-compdocs/generators/doc.go
@@ -302,7 +302,7 @@ func flagUsages(f *pflag.FlagSet) string {
 			line += fmt.Sprintf("--%s", flag.Name)
 		}
 
-		varname, usage := UnquoteUsage(flag)
+		varname, usage := unquoteUsage(flag)
 		if len(varname) > 0 {
 			line += " " + varname
 		}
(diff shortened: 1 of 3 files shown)


kbhawkey:
Hi @tengqm .
I read over the changes and looked at the CI build. There is new build output?
Is this correct?

tengqm:
Yes. This PR changes the data type presentation so that readers will see `<comma-separated 'key=True|False' pairs>` instead of `mapStringBool`.

gen-compdocs/generators/doc.go
```diff
@@ -412,13 +418,33 @@ func UnquoteUsage(flag *pflag.Flag) (name string, usage string) {
...
+		name = "string"
+	case "moduleSpec":
+		name = "&lt;comma-separated 'pattern=N' settings&gt;"
+	case "portRange":
+		name = "&lt;a string in the form 'N1-N2'&gt;"
+	case "traceLocation":
+		name = "&lt;a string in the form 'file:N'&gt;"
 	}
-
+	fmt.Printf("type name = %s\n", name)
```
kbhawkey, line 448: Do you want to keep this line?
https://travis-ci.org/github/kubernetes-sigs/reference-docs/builds/760101045

kbhawkey:
Thanks @tengqm .
/lgtm

PR #215, opened 2021-03-06
Add resourcequota config

tengqm:
This adds the support to generate resourcequota configuration reference. We also prefer the GroupName used over the 'groupName' comment which sometimes deviates from the implementation.

related: https://github.com/kubernetes/website/pull/26820

### DIFF
--- a/genref/config.yaml
+++ b/genref/config.yaml
@@ -116,6 +116,11 @@ apis:
     package: k8s.io/apiserver
     path: pkg/apis/audit/v1
 
+  - name: apiserver-resourcequota
+    title: kube-apiserver ResourceQuota Configuration (v1)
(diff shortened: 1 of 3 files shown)


tengqm:
Holding this pending #176
/hold
/approve cancel

genref/config.yaml
```diff
@@ -116,6 +116,11 @@ apis:
     package: k8s.io/apiserver
     path: pkg/apis/audit/v1
 
+  - name: apiserver-resourcequota
+    title: kube-apiserver ResourceQuota Config (v1)
```
kbhawkey, line 120: nit: The other title strings include "Configuration" not "Config".

tengqm: Fixed.

PR #216, opened 2021-03-10
Fix bug in reference generation

tengqm:
The bug was introduced in last commit where we added new data type representation.


kbhawkey:
Thanks @tengqm .
/lgtm

PR #218, opened 2021-03-25
Improve comp reference generator

tengqm:
This PR ensures that included sub-command references also carry the "auto-generated" warning.


irvifa (commented):
/lgtm

PR #219, opened 2021-03-26
Update API reference generator

tengqm:
This PR updates the config for 1.20 release and also adds go.mod for
running the generator.


kbhawkey:
@tengqm This is for v1.21?

tengqm:
@kbhawkey This is version agnostic. It is about using 'go.mod' for API reference generation.

PR #220, opened 2021-04-01
Use code tag instead of samp for config API types

tengqm:
This PR optimizes the config API generator and fixed some nits:
    
- Use code tag instead of samp for config API types (team agreed to fix `<code>` style rather than using `<samp>`)
- Skip more packages because they are not used, buggy, or superceded
- Improved `pkg` template logic for testing group name
- Added detection for duplicated entries in references, which was identified during review

### DIFF
--- a/genref/config.yaml
+++ b/genref/config.yaml
@@ -38,6 +38,8 @@ apis:
     title: Kubelet Stats (v1alpha1)
     package: k8s.io/kubelet
     path: pkg/apis/stats/v1alpha1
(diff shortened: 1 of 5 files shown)


kbhawkey:
👀

kbhawkey:
@tengqm , Do you want to add a class attribute for the `code` elements?
Changes look good.

/lgtm

Hold to check on code elements, otherwise feel free to remove the hold. Thanks!
/hold

PR #224, opened 2021-04-09
Update reference generator for kubectl

tengqm:
This PR migrates the `kubectl` reference generator to use go modules. Since the migration of `kubectl` into staging is almost done, the generator doesn't need to import the `k8s.io/kubernetes` repo.


kbhawkey:
Hold for review.
The changes seem OK.
With these changes, I am wondering if it is possible to use the `k8s/website` update-imported-docs script?
If not, what problems need fixing?
/hold

tengqm:
@kbhawkey This PR only focus on simplifying the generators and adding more targets to CI jobs. I'll look at the website script and propose fixes if necessary.

tengqm:
@kbhawkey Nothing needs to change from the website scripts for kubectl. This fix focuses on simplifying the generator.

gen-kubectldocs/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/kubernetes-sigs/reference-docs/gen-kubectldocs
+
+go 1.15
```
kbhawkey, line 3: nit: go 1.16?

tengqm: Yup. 1.16 would be better.

tengqm: Fixed.

kbhawkey:
Sorry for the delay in reviewing.
I am assuming these changes are up-to-date.
/lgtm

PR #227, opened 2021-04-10
Fix docsy example

### LINKED ISSUE #226: It is not possible to use this repository as a git submodule
It is not possible to use this repository as a git submodule.

Steps to reproduce:

```
$ mkdir test
$ cd test
$ git init
$ git submodule add https://github.com/kubernetes-sigs/reference-docs.git
$ git submodule update --init --recursive 
fatal: No url found for submodule path 'reference-docs/docsy-example/themes/docsy' in .gitmodules
Failed to recurse into submodule path 'reference-docs'
```

I can see there is a .gitmodules file in the `docsy-example` directory, but I would suggest it should be in the root directory.

kbhawkey:
I previously added the docsy-example site in this repository under the docsy-example directory.
This PR moves that content to the root of this repository and updates the theme.
Closes #226


kbhawkey:
I could remove most of the "content" from the example site.
/cc @tengqm

tengqm:
Let's kick this in and iterate over.
/approve

PR #231, opened 2021-04-30
Fix component reference generator

tengqm:
The upstream source code now applies normalization function for all
flags. This PR syncs such a behavior to reference generator.


kbhawkey:
Hi @tengqm .
Thanks for the updates.
You have generated/tested the output?
/lgtm
You can unhold if the generated flags are corrected.
/hold

PR #234, opened 2021-05-11
Fix reference generator for kubeadm

tengqm:


kbhawkey:
Thanks @tengqm .
/lgtm

PR #238, opened 2021-07-03
Fix doc strings for kubeadm config

tengqm:
There are some strange format in the kubeadm comment that led to weird
output for the generated reference. This PR fixes them.

### DIFF
--- a/genref/go.mod
+++ b/genref/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/kr/text v0.2.0 // indirect
 	github.com/niemeyer/pretty v0.0.0-20200227124842-a10e7caefd8e // indirect
 	github.com/pkg/errors v0.9.1
-	github.com/tengqm/kubeconfig v0.0.0-20210622015104-4c2594d38f88 // indirect
+	github.com/tengqm/kubeconfig v0.0.0-20210703133455-61479ab64f9d
 	github.com/yuin/goldmark v1.2.1
 	github.com/yuin/goldmark-highlighting v0.0.0-20200307114337-60d527fdb691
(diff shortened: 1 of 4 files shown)


sftim:
I'm happy to review this but don't actually know the steps to check it's accurate.

Do we have a guide to that?

tengqm:
@sftim This is a trivial update to existing code base. The kubeadm config is deeply embedded in the Kubernetes/kubernetes code base, but k8s.io/kubernetes cannot be imported directly for parsing. So the workaround, as suggested by some contributors was to clone the upstream source (especially the `types.go`) files for parsing. For that reason, I cloned the related source code into github.com/tengqm/kubeconfig repo and then parse it.

Regarding this PR, it bumps the version of the `github.com/tengqm/kubeconfig` module. In the latest revision, we further cleaned the `doc.go` file where there are bad TABs found in the comments. Such comments, as @kbhawkey has spotted, are leading us to incorrect sample YAML data.

kbhawkey:
Hi @tengqm . I assume that this PR updates the `genref` tool and the generated configuration reference pages. The changes are based on files found in github.com/tengqm/kubeconfig:
https://github.com/tengqm/kubeconfig/blob/master/config/kubeadm/v1beta1/doc.go
https://github.com/tengqm/kubeconfig/blob/master/config/kubeadm/v1beta2/doc.go

tengqm:
@kbhawkey Correct.

kbhawkey:
Hi. The formatting changes look okay.
Does the tool need more detailed documentation? It would be helpful to explain any dependencies or known issues with upstream source/libraries. What workarounds are used to build the configuration references?
Thanks @tengqm .
/lgtm

PR #247, opened 2021-07-29
Update latest events API to events.k8s.io

dgrisonnet:
In Kubernetes v1.19, the new Events API events.k8s.io was promoted to
v1. As such it now supersedes the original core Events API. Since this
wasn't defined in this project, the official kubernetes-api doc was
putting the events.k8s.io API under the list of old APIs and the
recommended API for Events was still the core one. To fix that, we need
to define that the events API group is newer than the core one.

/cc @ehashman

### DIFF
--- a/gen-apidocs/config/config.yaml
+++ b/gen-apidocs/config/config.yaml
@@ -119,7 +119,7 @@ resource_categories:
       group: apiextensions
     - name: Event
       version: v1
-      group: core
+      group: events
     - name: LimitRange
       version: v1
       group: core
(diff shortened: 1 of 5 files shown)


dgrisonnet:
I was trying to regenerate the website pages based on this change and everything went fine for v1.20 and v1.21, but I ran into some errors while running `make copyapi` for v1.19:

```console
Warning: full name for 'batch' not provided, guessing...
...
No Definition found for listRbacAuthorizationV1RoleForAllNamespaces
```

Are the instructions to generate the API doc different from the more recent versions?

/hold

dgrisonnet:
@feloy @tengqm could you please have a look?

tengqm:
/lgtm
/approve
/hold cancel
I'll look into the v1.19 issue separately.

PR #249, opened 2021-08-05
Update go mod for component ref for 1.22

tengqm:


PI-Victor:
ah, there it is, i was wondering how these are updated. can you explain in a doc, somewhere on the main README or anywhere else what the process is for updating the deps? 

This would save sig-release a lot of time. I tried all sorts of solutions, but nothing worked since i had no idea what are they supposed to point to.

sftim:
No objection, but I'm not sure how to verify that this is the right fix.

kbhawkey:
Thanks @tengqm .
/lgtm

tengqm:
@PI-Victor I've filed an issue on this in case the ball gets dropped. I'm in the middle of some activities. Will get back to this when time permits.

PR #264, opened 2021-12-08
Update go module for 1.23 component reference

tengqm:


sftim:
I'm afraid I don't know how to verify if this change is appropriate / correct.

kbhawkey:
Hi @tengqm .
Should the component pages build with Go 1.17?
I built several kubeadm files with the component build? Is this correct? Thanks.

kbhawkey:
I used these updates to build the component tool pages (not sure if correct):
https://github.com/kbhawkey/reference-docs/blob/0565d9c642284907fe577530d145071b1f86af74/gen-compdocs/go.mod

tengqm:
@kbhawkey I'm not sure we need to include a complete list of indirectly depended packages in go.mod. Usually I'd only add whatever package we reference directly in the Go source and leave the rest to the go mod system. Additional packages may be needed if go build is unhappy with the package list we provided.

PR #265, opened 2021-12-08
Configuration API for v1.23

tengqm:


genref/go.mod
```diff
@@ -3,31 +3,33 @@ module github.com/kubernetes-sigs/reference-docs/genref
 go 1.16
```
kbhawkey, line 3: Should you build with Go 1.17?

genref/config.yaml
```diff
@@ -44,18 +44,29 @@ apis:
     # Skip this because not referenced anywhere and it has no group name
     skip: true
 
+  - name: kubelet-credentialprovider
+    title: Kubelet CredentialProvider (v1alpha1)
```
kbhawkey, line 48: At release time, how do you know what to add to config.yaml?

genref/go.mod
tengqm: Tried, but failed. Intra-package and inter-package references are not correctly resolved in 1.17. Need some time to investigate.

genref/config.yaml
tengqm: I'm not aware of any solution that can automate the discovery of new versions.

tengqm:
/hold
Reference target is not correct.

genref/config.yaml
kbhawkey: OK. I am trying to sort out how we can provide instructions for the release team (and others) to generate these reference pages.

kbhawkey:
These changes look ready to merge. Thanks @tengqm .
/lgtm
Feel free to release the hold.
/hold

tengqm:
Releasing the hold for follow up work.
/hold cancel

PR #267, opened 2021-12-09
remove docsy theme

kbhawkey:
The theme was added so that it would make it easier to test the generated files.
However, the theme has not been used all that much. This PR removes the files associated with the theme.


tengqm:
/lgtm
Thanks for the clean up.

PR #269, opened 2021-12-15
Fix nits in the config API generator

tengqm:
This PR updates the templates so that we are outputing references only for the relevant types. This is to fix a nit where in some outputs (e.g. kube-proxy-config), irrelevant type definitions got sneaked in. A better solution would be to export the resource group name into the configuration file. However, that would be trickier to maintain than this workaround.

### DIFF
--- a/genref/output/md/apiserver-audit.v1.md
+++ b/genref/output/md/apiserver-audit.v1.md
@@ -16,18 +16,16 @@ auto_generated: true
   
     
 
-
 ## `Event`     {#audit-k8s-io-v1-Event}
     
-
-
 
 **Appears in:**
 
 - [EventList](#audit-k8s-io-v1-EventList)
 
 
-Event captures all the information that can be included in an API audit log.
+<p>Event captures all the information that can be included in an API audit log.</p>
+
 
 <table class="table">
 <thead><tr><th width="30%">Field</th><th>Description</th></tr></thead>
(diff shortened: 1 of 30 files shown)


kbhawkey:
👀

kbhawkey:
👀

genref/output/md/apiserver-audit.v1.md
```diff
@@ -248,390 +214,323 @@ categories are logged.
...
 'pods/log' matches the log subresource of pods.
-'&lowast;' matches all resources and their subresources.
-'pods/&lowast;' matches all subresources of pods.
-'&lowast;/scale' matches all scale subresources.
-
-If wildcard is present, the validation rule will ensure resources do not
-overlap with each other.
-
-An empty list implies all resources and subresources in this API groups apply.</td>
+'<em>' matches all resources and their subresources.
```
kbhawkey, line 320: nit:  Is the quoted `<em>` element correct?

genref/output/md/cloud-provider-service.v1alpha1.md
```diff
@@ -9,32 +9,29 @@
...
 <tbody>
     
-
   
 <tr><td><code>ConcurrentServiceSyncs</code> <B>[Required]</B><br/>
 <code>int32</code>
 </td>
 <td>
-   concurrentServiceSyncs is the number of services that are
+   <p>concurrentServiceSyncs is the number of services that are
```
kbhawkey, line 31: nit: Just checking if the field name is `concurrentServiceSyncs` or `ConcurrentServiceSyncs`?

kbhawkey:
Reviewed Go package name updates, minor HTML formatting changes, and checks to only display relevant types.
/lgtm

genref/output/md/cloud-provider-service.v1alpha1.md
tengqm: This is a good question. I have to check the upstream source code to give you an answer ...

The source code corresponding to the field is located [here](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cloud-provider/controllers/service/config/v1alpha1/types.go#L19-L25). As you can see from the snippet copied below, the struct definition failed to include a JSON tag for the member field `ConcurrentServiceSyncs`.

```go
// ServiceControllerConfiguration contains elements describing ServiceController.
type ServiceControllerConfiguration struct {
	// concurrentServiceSyncs is the number of services that are
	// allowed to sync concurrently. Larger number = more responsive service
	// management, but more CPU (and network) load.
	ConcurrentServiceSyncs int32
}
```

In other words, the declaration of the field should be something like (**note how the user facing field name is specified using this tag**):

```go
     ConcurrentServiceSyncs int32  `json: "concurrentServiceSyncs"`
```

Without this tag, the configuration will have to contain "`ConcurrentServiceSyncs`" for it to be valid. However, the comment string is still putting it as 'concurrentServiceSyncs'. So this smells like bug in the source code.

genref/output/md/apiserver-audit.v1.md
tengqm: This is incorrect. We still have some nits in parsing the comment string. The upstream [source code](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/apis/audit/v1/types.go#L257-L271) looks like this:

```go
	// Resources is a list of resources this rule applies to.
	//
	// For example:
	// 'pods' matches pods.
	// 'pods/log' matches the log subresource of pods.
	// '*' matches all resources and their subresources.
	// 'pods/*' matches all subresources of pods.
	// '*/scale' matches all scale subresources.
	//
	// If wildcard is present, the validation rule will ensure resources do not
	// overlap with each other.
	//
	// An empty list implies all resources and subresources in this API groups apply.
	// +optional
	Resources []string `json:"resources,omitempty" protobuf:"bytes,2,rep,name=resources"`
```

The problems is that `*` in the comments are sometimes used as a markdown tag, while sometimes used as a literal string. To make things consistent, we will need some protocols for the whole development community.

I'll try if escaping the '*' character before handling the comment to the goldmark will yield better results.

kbhawkey: 👍

genref/output/md/cloud-provider-service.v1alpha1.md
kbhawkey: 👍

PR #272, opened 2022-03-11
Fix doc strings for kubeadm-config v1beta3

tengqm:


genref/output/md/kubeadm-config.v1beta3.md
```diff
@@ -264,6 +264,109 @@ node only (e.g. the node ip).</p>
...
+## `BootstrapTokenString`     {#BootstrapTokenString}
+    
+
+**Appears in:**
+
+- [BootstrapToken](#BootstrapToken)
+
+
+<p>BootstrapTokenString is a token of the format <code>abcdef.abcdef0123456789</code> that is used
+for both validation of the practically of the API server from a joining node's point
```
kbhawkey, line 343: nit: Is this content taken from the source? I don't understand the phrase, `for both validation of the practically of the API server `

tengqm: Yes. It is from the source: https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/apis/bootstraptoken/v1/types.go#L51-L54

kbhawkey: Okay. We should improve the descriptions found in the source.
/lgtm

PR #275, opened 2022-05-15
Update kubectl generator for 1.24

liggitt:
xref https://github.com/kubernetes/website/pull/33676 and https://github.com/kubernetes/website/issues/33674


gen-kubectldocs/generators/gen.go
```diff
@@ -252,6 +252,7 @@ func WriteCommandFile(manifest *Manifest, t *template.Template, params TopLevelC
 		">", "&gt;",
 		"[", "<span>[</span>",
 		"]", "<span>]</span>",
+		"\n", "<br>",
```
liggitt, line 255: generator choked on usage strings that included newlines

tengqm:
/approve
Thanks.

sftim:
/approve

I'm not much cop with Golang so won't LGTM.

PR #276, opened 2022-05-17
Add eventratelimit config API v1alpha1

tengqm:
The package was in k8s.io/kubernetes which cannot be directly imported.
I've copied that file into github.com/tengqm/kubeconfig/admission for
parsing.


kbhawkey:
👍 
/lgtm

PR #281, opened 2022-06-05
Fix component reference generator

tengqm:
There are some command line arguments (e.g. `default-not-ready-toleration-seconds`) registered for the
`kube-apiserver` as GLOBAL flags. Such kind of flag registration is somehow impacting the `kube-proxy` command line flags. This PR separates the proxy command parsing into a separate main file so that API server args are not impacting the kube-proxy module.


gen-compdocs/comps/gen_kube_docs.go
```diff
@@ -45,27 +45,23 @@ func GenerateFiles(path, module string) {
...
+		generators.GenMarkdownTree(apiserver, outDir, true)
 
 	case "kube-controller-manager":
 		pflag.CommandLine.SetNormalizeFunc(cliflag.WordSepNormalizeFunc)
 		controllermanager := cmapp.NewControllerManagerCommand()
-		GenMarkdownTree(controllermanager, outDir, true)
+		generators.GenMarkdownTree(controllermanager, outDir, true)
 
 	case "kube-proxy":
-		proxy := proxyapp.NewProxyCommand()
```
kbhawkey, line 56: What is supposed to happen in the `kube-proxy` case?

tengqm: good catch.

kbhawkey:
Thanks @tengqm .
/lgtm

PR #292, opened 2022-08-24
Update configuration for API reference 1.25

tengqm:


kbhawkey:
Thanks @tengqm .
/lgtm

PR #300, opened 2022-09-21
Update configuration YAML for API reference v1.25

tengqm:


kbhawkey:
Thanks @tengqm .
/lgtm

PR #303, opened 2022-12-08
Update configuration for 1.26 API reference

tengqm:
There are new resource groups and resources introduced in 1.26. This PR updates the configuration for API reference generation.


tengqm:
/hold

Depends on #302

PR #321, opened 2023-05-18
Improve generation of config APIs

tengqm:
Update the kube-controller-manager config API by fixing its API group. Added two more metrics APIs.


sftim (commented):
I prefer to split changes into:
- generator change
- the outcome of that generator (second commit)

this makes reverts and merges a bit easier if we need them.

/approve

PR #324, opened 2023-07-26
Support referenced parameters

feloy:
Kubernetes v1.28 spec introduces the use of referenced Parameters:

```
        "parameters": [
          {
-            "in": "body",
-            "name": "body",
-            "schema": {
-              "$ref": "#/definitions/io.k8s.apimachinery.pkg.apis.meta.v1.DeleteOptions"
+            "$ref": "#/parameters/body-2Y1dVQaQ"
            }
          },
```

The changes in this PR makes the converter able to get the parameter information from the referenced parameter.

This change is necessary to build the multi-pages API reference for Kubernetes v1.28

### DIFF
--- a/gen-resourcesdocs/pkg/config/output_test.go
+++ b/gen-resourcesdocs/pkg/config/output_test.go
@@ -36,7 +36,7 @@ func (o FakeSection) AddContent(s string) error                              { r
 func (o FakeSection) AddTypeDefinition(typ string, description string) error { return nil }
 func (o FakeSection) AddFieldCategory(name string) error                     { return nil }
 
-func (o FakeSection) AddProperty(name string, property *kubernetes.Property, linkend []string, indent bool, defname string, shortName string) error {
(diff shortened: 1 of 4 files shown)


gen-resourcesdocs/pkg/config/output_test.go
```diff
@@ -36,7 +36,7 @@ func (o FakeSection) AddContent(s string) error                              { r
 func (o FakeSection) AddTypeDefinition(typ string, description string) error { return nil }
 func (o FakeSection) AddFieldCategory(name string) error                     { return nil }
 
-func (o FakeSection) AddProperty(name string, property *kubernetes.Property, linkend []string, indent bool, defname string, shortName string) error {
+func (o FakeSection) AddProperty(name string, property *kubernetes.Property, linkend []string, indent int, defname string, shortName string) error {
```
feloy, line 39: This change is not related to the main change, but is necessary to make all the tests pass

PR #325, opened 2023-08-11
Update API reference generator for 1.28

tengqm:


sftim:
Could you explain the change to `initOperationParameters()` @tengqm? It looks reasonable but I don't have context for why we're updating that function here.

tengqm:
> Could you explain the change to `initOperationParameters()` @tengqm? It looks reasonable but I don't have context for why we're updating that function here.

The new swagger doc starting v1.28 has some drastical changes to how path parameters are documented. Previously, the swagger generator (in the upstream source code) generates the complete list of parameters for each and every operation, and the parameters are repeatedly documented underneath that operation. For example, every "listNamespacedFooBar" operation has a `watch` parameter documented. This means the how swagger spec is large, with duplicated contents.

Starting with v1.28, the generator was "optimized". All parameters are extracted to the top level swagger `parameters` block, leaving the per-operation parameter definition merely a reference to the global one. The has heavily reduced the spec size. It was a good intent.

Our previous code wasn't able to handle this case. So we have to parse the parameter reference, fetch the global parameter definitions and document them on a per-operation basis.

There are significant bugs in the generated swagger today. Mainly because the logic cannot handle duplicated parameter names properly. For example, for `connectNamespacedPod....ProxyWithPath` operations, there are two parameters named `path`, one in the URL path, the other in the body. The generated swagger in 1.28 is not handling this correctly. That is something to be fixed in the upstream.

sftim:
> There are significant bugs in the generated swagger today. Mainly because the logic cannot handle duplicated parameter names properly. For example, for `connectNamespacedPod....ProxyWithPath` operations, there are two parameters named path, one in the URL path, the other in the body. The generated swagger in 1.28 is not handling this correctly. That is something to be fixed in the upstream.

Is there an issue? Let's try to file one.

/lgtm

PR #339, opened 2023-09-21
Update pkg dependency for genref

tengqm:


xrstf:
@tengqm Thanks a lot for your super quick reviews on all these PRs, it's really appreciated :-) Is there anything else I need to do to complete this? The markup on https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/ is still the old one, but the new CSS/JS is already live.

PR #345, opened 2023-12-05
Makefile edits for containerization of generation script

Gauravpadam:
This PR is associated with https://github.com/kubernetes/website/pull/44029

The aim is to containerize the script which is generating these docs
These changes are well tested on my personal fork and they work perfectly fine

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@
 WEBROOT=${K8S_WEBROOT}
 K8SROOT=${K8S_ROOT}
 K8SRELEASE=${K8S_RELEASE}
+INSIDE_CONTAINER=${CONTAINER}
 
 ifeq ($(K8SRELEASE),)
   $(error Please define K8S_RELEASE, e.g. 'export K8S_RELEASE=1.21.0')
@@ -44,8 +45,14 @@ cleancli:
 
 cli: cleancli
 	cd gen-kubectldocs && go mod download && go run main.go --kubernetes-version v$(K8SRELEASEDIR)
-	mkdir -p $(CLISRC)
-	docker run -v $(shell pwd)/gen-kubectldocs/generators/includes:/source -v $(shell pwd)/gen-kubectldocs/generators/build:/build -v $(shell pwd)/gen-kubectldocs/generators/:/manifest brianpursley/brodocs:latest
+	mkdir -p $(CLISRC) && \
+	if [ "$(INSIDE_CONTAINER)" = "True" ]; then\
+		cd /app/brodocs/brodocs && ./runcontainerdocs.sh;\
+	else\
+		docker run -v "$(shell pwd)/gen-kubectldocs/generators/includes:/source" -v "$(shell pwd)/gen-kubectldocs/generators/build:/build" -v "$(shell pwd)/gen-kubectldocs/generators/:/manifest" brianpursley/brodocs:latest;\
+	fi
+
+
 
 copycli: cli
 	cp gen-kubectldocs/generators/build/index.html $(WEBROOT)/static/docs/reference/generated/kubectl/kubectl-commands.html


Makefile
```diff
@@ -44,8 +45,14 @@ cleancli:
...
 cli: cleancli
 	cd gen-kubectldocs && go mod download && go run main.go --kubernetes-version v$(K8SRELEASEDIR)
-	mkdir -p $(CLISRC)
-	docker run -v $(shell pwd)/gen-kubectldocs/generators/includes:/source -v $(shell pwd)/gen-kubectldocs/generators/build:/build -v $(shell pwd)/gen-kubectldocs/generators/:/manifest brianpursley/brodocs:latest
+	mkdir -p $(CLISRC) && \
+	if [ "$(INSIDE_CONTAINER)" = "True" ]; then\
+		cd /app/brodocs/brodocs && ./runcontainerdocs.sh;\
+	else\
+		docker run -v "$(shell pwd)/gen-kubectldocs/generators/includes:/source" -v "$(shell pwd)/gen-kubectldocs/generators/build:/build" -v "$(shell pwd)/gen-kubectldocs/generators/:/manifest" brianpursley/brodocs:latest;\
+	fi
```
tengqm, line 53: I'm not quite sure that we are heading this direction. We have suffered quite some from running the containerized generator for kubectl docs. We were supposed to adopt the generator into k8s website project, but we didn't do that. So this is always a 3rd-party dependency. We cannot update the source code even when we have identified bugs in generating kubectl docs lately.  We need a registry for the container image, accessible to the maintainers who can commit to maintain it for a long time. By maintenance, I mean we need to ensure at least the latest version works, and there is no significant CVEs in it.

So, tell me, my friend. What does containerization buy us in this case? I am not saying that it is not doable. Actually, we can make it work. What I am curious is about the benefits of containerizing the generator. How many people around the world will regenerate the references frequently? Who will be maintaining the container images at the end of the day?

Gauravpadam: This is insightful and sad at the same time. I know it has been a hassle for us to maintain reference documentation and it's generation. But I believe containerizing might open up doors to future improvements

At the moment, we have to rely on third party images. What we can do is, create an image repository and take it under the Kubernetes project. That way, Our maintainers can maintain the images. I'm here to stay and I'll be happy to maintain my version of `gauravpadam/brodocs:latest` for a long time

This is not a partial approach which only containerizes Kubectl generation, It does so for the whole script. There are some errors and deprecations on a regular basis but we can address them if we truly have the incentive. Kubernetes project being so vast, only has < 5000 contributors as of now, Active contributors? < 1000, Maybe even less. That's just the way it is. I think instead of considering who will generate the references, We should put one foot forward at a time and see where this goes. I'll be happy to assist in documentation generation wherever required

tengqm: Glad to know that there are still enthusiasts passionate on making all these things better. What I was trying to say is that containerization may be an improvement in and of itself, but the maintenance overhead over-weighs the benefit we can expect from it.
If you are really into this, I'd suggest you turn to some more meaningful, useful improvements. For example, the component reference generation has been failing for kubelet for quite some time. We need to investigate how to get those command line flags exposed again. Manually editing the kubelet reference has been a pain for quite a few cycles. We may need to improve our API reference generator to use OpenAPI v3, which is somehow deprecating the old swagger one. We may want to give up the current brodocs way of kubectl reference because the look and feel is so different from the rest of the k8s website. The examples in the reference are not that interesting as we believed. We can generate kubectl reference using the `gen-compdocs` tool, and we have proposed a replacement. It was not accepted because someone wants the files to be placed in a better hierarchy. So on and so forth.

Gauravpadam: Let's make it happen. I don't mind keeping this proposal open till we can actually merge it without overhead. I'll start by looking more into the actual problems faced by our maintainers. Hopefully we'll come up with a streamlined solution which fits all our criteria

kbhawkey:
Hello @Gauravpadam .
I'm not clear that this solves the generation and maintenance of the kubectl reference docs.

Gauravpadam:
This fixes https://github.com/kubernetes/website/issues/28037, as an extension of https://github.com/kubernetes/website/pull/44029

sftim:
@Gauravpadam could you write up an explanation of how to:
- test that the target branch is not right
- test that these changes fix that issue

?

That advice will help people to review the change.

Gauravpadam:
Sure @sftim, I already have some documented steps open as a PR if that might help;
https://github.com/kubernetes/website/pull/44055

I'll also be happy to assist in any way to make the reviewing process simpler

Gauravpadam:
You were a good pr

/close

PR #347, opened 2023-12-14
Update configuration for API ref 1.29

tengqm:


kbhawkey:
I read through the changes and assume that you have verified the generation of the docs with these updates.
/lgtm

tengqm:
> I read through the changes and assume that you have verified the generation of the docs with these updates. /lgtm

Yes I did. Thanks for review.

PR #348, opened 2023-12-14
Update configuration for component reference v1.29

tengqm:


gen-compdocs/generators/doc.go
sftim, line 1: nit: I'd split this into its own commit.

sftim (commented):
/lgtm

Thanks

PR #349, opened 2024-01-06
Improve reference generator for kubectl

tengqm:
Improve component reference generator for kubectl so that the references are generated into sub-directories.
This change has no impact on other components like kube-apiserver, kube-controller-manager.
It make impact how kubeadm reference are generated.

### DIFF
--- a/gen-compdocs/generators/doc.go
+++ b/gen-compdocs/generators/doc.go
@@ -54,21 +54,41 @@ guide. You can file document formatting bugs against the
 func GenMarkdownTree(cmd *cobra.Command, dir string, withTitle bool) error {
 	identity := func(s string) string { return s }
 	emptyStr := func(s string) string { return "" }
-	return GenMarkdownTreeCustom(cmd, dir, emptyStr, identity, withTitle)
+	return GenMarkdownTreeCustom(cmd, dir, "", emptyStr, identity, withTitle)
 }
 
-func GenMarkdownTreeCustom(cmd *cobra.Command, dir string, filePrepender, linkHandler func(string) string, withTitle bool) error {
+func GenMarkdownTreeCustom(cmd *cobra.Command, dir string, subdir string, filePrepender, linkHandler func(string) string, withTitle bool) error {
 	for _, c := range cmd.Commands() {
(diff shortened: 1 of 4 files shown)


sftim:
>  It make impact how kubeadm reference are generated.

Can you explain more about that @tengqm ? If there might be an impact, it'll help to know what to look for (as a reviewer).

tengqm:
> > It make impact how kubeadm reference are generated.
> 
> Can you explain more about that @tengqm ? If there might be an impact, it'll help to know what to look for (as a reviewer).

The logic introduced will have an impact on all kube components that have sub-commands. This is in general not an issue for kube-apiserver or kube-proxy etc. But for kubeadm, there are sub commands. The generated files will be nested into subdirectories as well. Need to investigate.

gen-compdocs/generators/doc.go
```diff
@@ -54,21 +54,41 @@ guide. You can file document formatting bugs against the
...
 	for _, c := range cmd.Commands() {
 		if !c.IsAvailableCommand() || c.IsAdditionalHelpTopicCommand() {
 			continue
 		}
-		if err := GenMarkdownTreeCustom(c, dir, filePrepender, linkHandler, withTitle); err != nil {
+		// CommandPath example: 'kubectl top pod'
+		parts := strings.Split(c.CommandPath(), " ")
+		subdir := ""
+		if len(parts) > 1 {
+			subdir = parts[0] + "_" + parts[1]
```
stmcginnis, line 69: In case there are more parts, would it be better to do:

```go
subdir = strings.Join(parts, "_")
```

I can't think of a specific case right now, but I believe it's possible some commands will have more parts.

sftim: Maybe we don't want the tree structure to be too deep?
(a comment explaining intent would help a lot here)

sftim:
> The logic introduced will have an impact on all kube components that have sub-commands. This is in general not an issue for kube-apiserver or kube-proxy etc. But for kubeadm, there are sub commands. The generated files will be nested into subdirectories as well. Need to investigate.

Maybe we'll want redirects if we do change the URLs vs. existing `kubeadm` docs. If we do, adding those is a one-off task.

PR #351, opened 2024-02-01
fix: sort types

### LINKED ISSUE #350: Sort types is broken
Sorting types only by name is broken if multiple types have the same name.
We should consider the package too.

https://github.com/kubernetes-sigs/reference-docs/blob/aed673f90d888570c463cb9bff99e1c7d0fd66da/genref/types.go#L406

eddycharly:
This PR fixes `sortTypes` func by considering the package when sorting, in case both types have the same name.

Let's say a type `Foo` exists in `v1` and `v2`.
If we consider only the name (`Foo`) we can't decide on the order, it will be random from one run to the other.
In this case, we can additionally consider the package, it becomes obvious that `v1` should come before `v2` and the order is now predictible.

Fixes #350

### DIFF
--- a/genref/types.go
+++ b/genref/types.go
@@ -403,7 +403,10 @@ func sortTypes(typs []*apiType) []*apiType {
 		} else if !t1.IsExported() && t2.IsExported() {
 			return false
 		}
-		return t1.Name.Name < t2.Name.Name
+		if t1.Name.Name != t2.Name.Name {
+			return t1.Name.Name < t2.Name.Name
+		}
+		return t1.Name.String() < t2.Name.String()
 	})
 	return typs
 }


sftim (commented):
Thank you for the PR.

I'd split this into 2 commits:
- update to the generator code
- the consequences of that update

---

Please add a more detailed explanation to the commit that changes the generator code.

eddycharly:
@sftim i added a more detailed explanation, is it more clear ?

sftim (commented):
Thanks

/lgtm

eddycharly:
🤣

tengqm (approved):
/approve

PR #352, opened 2024-03-15
fix: show example button with bootstrap 5

GitBluub:
Fixing the show example button in the reference docs not using the correct data attributes for bootstrap v5 https://getbootstrap.com/docs/5.0/components/collapse/#via-data-attributes


tengqm:
Thanks.
/lgtm
/approve

PR #354, opened 2024-03-22
Add pod/exec and pod/attach to api reference

dshebib:
Add exec and attach operations to the generated API reference documentation.


tengqm:
This is clean and good.
/lgtm
/approve

PR #358, opened 2024-04-19
Update config and buffered data for configapi 1.30

tengqm:


reylejano:
Thank you for the version bump to 1.30
/lgtm

PR #362, opened 2024-05-13
Adds Kat Cosgrove as SIG Docs tech lead

katcosgrove:
Adding myself as SIG Docs tech lead!


salaxander:
@katcosgrove looks like we need to get you added to the Kubernetes-sigs org for this one

tengqm:
How do we proceed with this?

PR #363, opened 2024-06-05
Create section for non-resource-associated endpoints

dshebib:
This PR adds a section to the generated API reference containing operations which are not associated with a particular resource.

PR #364, opened 2024-06-05
Add special case operation names

dshebib:
Adds a few more operations to the api documentation in the appropriate locations.

### DIFF
--- a/gen-apidocs/config/v1_30/config.yaml
+++ b/gen-apidocs/config/v1_30/config.yaml
@@ -309,6 +309,10 @@ operation_categories:
       match: proxy${group}${version}PUT(Namespaced)?${resource}
     - name: Replace Proxy Path
       match: proxy${group}${version}PUT(Namespaced)?${resource}WithPath
+    - name: Patch Proxy
+      match: connect${group}${version}Patch(Namespaced)?${resource}Proxy
+    - name: Patch Proxy Path
+      match: connect${group}${version}Patch(Namespaced)?${resource}ProxyWithPath
   - name: "Misc Operations"
     default: true
     operation_types:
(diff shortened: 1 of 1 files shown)


gen-apidocs/config/v1_30/config.yaml
```diff
@@ -330,21 +334,18 @@ operation_categories:
...
-  - createCoreV1NamespacedPodBinding
-  - getCodeVersion
-  - logFileHandler
-  - logFileListHandler
-  - replaceCoreV1NamespaceFinalize
-  - V1beta1CertificateSigningRequestApproval
-  - V1CertificateSigningRequestApproval
-  - V1beta1NamespacedReplicationControllerDummyScale
-  - getServiceAccountIssuerOpenIDConfiguration
-  - getServiceAccountIssuerOpenIDKeyset
```
tengqm, line 349: Are these removals safe?

dshebib: The tool will emit log lines but those operations do not have an associated definition, so no documentation is generated for them and the documentation doesn't change. Feel free to verify.

tengqm: So ... this should be based on #363, right?

dshebib: I tested it both with and without #363. Without it, the endpoints that have new regex expressions are added and those that do not have a definition emit a message and do not get generated. With it, all previously excluded endpoints have documentation. It makes more sense after the previous commit but it wouldn't cause errors without it also be fine without it. I could merge them into a single PR if that makes more sense

gen-apidocs/config/v1_30/config.yaml
```diff
@@ -330,21 +334,18 @@ operation_categories:
...
+      match: replace${group}${version}(Namespaced)?${resource}Approval
 
 # List of *partial* operation IDs for matching. All matched operations are
 # excluded from the reference doc.
 excluded_operations:
-  - connectCoreV1Patch
-  - createCoreV1NamespacedPodBinding
-  - getCodeVersion
-  - logFileHandler
-  - logFileListHandler
```
tengqm, line 341: Please verify whether these operations are properly supported and thus should not be excluded.

dshebib: getCodeVersion: This seems to get the Kubernetes code version as intended
logFileHandle, logFileListHandler: Good catch, these are not supported

dshebib:
Should note that this PR will cause the generator to emit a few lines that are formatted as errors. @tengqm I am not familiar with the CI tools in this repo, should I reformat them as standard logs to avoid causing issues? The lines are the following:

```
[31mNo Definition found for getCodeVersion [/version/].[0m
[31mNo Definition found for getServiceAccountIssuerOpenIDKeyset [/openid/v1/jwks/].[0m
[31mNo Definition found for getServiceAccountIssuerOpenIDConfiguration [/.well-known/openid-configuration/].[0m
[31mNo Definition found for createCoreV1NamespacedPodBinding [/api/v1/namespaces/{namespace}/pods/{name}/binding].[0m
```

Emitted because though these operations have definitions, the generator does not associate them with a resource. The generated html after running the generator does not omit any pre-existing information and only moves information from the new Operations section to their respective sections (/pod/ etc.)

tengqm:
@dshebib The operations appear in the log because we don't have special logics to handle them.
In other words, our pattern matching logic cannot identify them as valid operations, and, that is the reason we temporarily excluded them from the reference docs generated.

For each operation from the swagger spec, we should either properly generate reference for that operator, or we should report it as an exception. We are not supposed to fail silently. This is the reason I asked you to double check this PR.

dshebib:
Sorry, I should have clarified. The documentation gets correctly generated according to the definition for the operation itself in the swagger.json. The error is generated because there is no associated _resource_ definition. Therefore, the final behaviour is that documentation is generated for the endpoint in a miscellaneous 'operations' section:

image
image

dshebib:
In checking this, I found that the logic from my previous PR does not respect and exclude the excluded operations in the config from the miscellaneous operations section. I will fix this this weekend.

PR #376, opened 2024-08-23
Add script to update api docs to latest released tags

dims:
Here's a script that fetches the latest released tags across the stable releases, updates the swagger json and generates new index.html files for all the new versions.

The idea is to to run this script from prow or github automation and then generate PR(s) to both this repository and the kubernetes/website repository automatically as a follow up.

kubernetes/website should be checked out and the environment variable K8S_WEBROOT must be set correctly before running the script. We use github API to fetch latest tags and directly pull swagger.json from github as well for specific tags.

### DIFF
--- a/hack/update-api-docs.sh
+++ b/hack/update-api-docs.sh
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+if [ ! -d "$K8S_WEBROOT" ]; then
+  echo "Error: missing $K8S_WEBROOT directory. Exiting..."
+  exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


tengqm:
I'm not quite sure we really want to do this automation.
We always need to adapt the configuration file when `1.<x+1>.0` is out, because there are usually noticeable changes to the API from `1.<x>` to `1.<x+1>`. Do we need to update and publish all the patch versions like `1.<x>.<y>`?

sftim:
This automation is not _necessary_, but it will help us to rehearse the API docs build ahead of the actual release - and that rehearsal is valuable.

PR #377, opened 2024-08-29
Update gen-resourcedocs for v1.31

tengqm:
This PR adds assets for generating markdown-style API reference for v1.31. It also fixes a problem related to the immature conversion of Markdown comments to HTML for resource descriptions.


dipesh-rawat (commented):
/lgtm

PR #378, opened 2024-09-30
Fix docs error for kubeadm config v1beta4

tengqm:
This PR updates the go module (tengqm/kubeconfig) for fixing the kubeadm config v1beta4 format errors.

### DIFF
--- a/genref/go.mod
+++ b/genref/go.mod
@@ -3,7 +3,7 @@ module github.com/kubernetes-sigs/reference-docs/genref
 go 1.22.0
 
 require (
-	github.com/tengqm/kubeconfig v0.0.0-20240819082452-df3693d84f6f
+	github.com/tengqm/kubeconfig v0.0.0-20240930134458-a0641b81f037
 	github.com/yuin/goldmark v1.7.4
 	github.com/yuin/goldmark-highlighting v0.0.0-20210516132338-9216f9c5aa01
 	k8s.io/api v0.31.0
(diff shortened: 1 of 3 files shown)


genref/go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/kubernetes-sigs/reference-docs/genref
 go 1.22.0
 
 require (
-	github.com/tengqm/kubeconfig v0.0.0-20240819082452-df3693d84f6f
+	github.com/tengqm/kubeconfig v0.0.0-20240930134458-a0641b81f037
```
sftim, line 6: I don't know how to check that this change is beneficial. Is there a good way to do that?

tengqm: The kubeadm config API lives in the k/k repo which cannot be imported by any other projects. In other words, it is under `cmd/kubeadm/app/...` directory rather than `staging/src/k8s.io/...`. That means the API definitions cannot be parsed from an external project. I created the `kubeconfig` project several years ago to solve this problem and copied (then sync'ed frequently) the API definitions, so that we can parse the `types.go` file and get the reference generated. In the `kubeconfig` project, the source code comments are largely revised for reference generation.

sftim: Could we move this repo, https://github.com/tengqm/kubeconfig, inside a Kubernetes org?

reylejano: I agree, can the repo https://github.com/tengqm/kubeconfig be inside a Kubernetes org

tengqm: Yes, I can transfer it to a kubernetes-sig for example.
What I am really interested in (and I did proposed to sig-cluster-lifcycle) is to move
the kubeadm API definition into `staging/src/k8s.io/kubeadm`.
If that happens, the whole community can import the APIs. Sigh ...

PR #379, opened 2024-10-01
feat: Improved documentation

### LINKED ISSUE #355: NetworkPolicy v1 is out of order in Cluster APIs section
Reference: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#networkpolicy-v1-networking-k8s-io

I'm looking at the v1.30 docs (I also saw this is the case for v1.28 and v1.29), and it looks like NetworkPolicy is out of place, as it's placed at the end, and the docs seem to be following an alphabetical order scheme. I'd adjust this, but I'm not sure if the files are generated from elsewhere nor how many versions are worth even addressing. Thanks.

Bharadwajshivam28:
fixes #355

### DIFF
--- a/gen-apidocs/config/config.yaml
+++ b/gen-apidocs/config/config.yaml
@@ -177,6 +177,9 @@ resource_categories:
     - name: Namespace
       version: v1
       group: core
+    - name: NetworkPolicy
+      version: v1
+      group: networking
     - name: Node
       version: v1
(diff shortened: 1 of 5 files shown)


Bharadwajshivam28:
@sftim @reylejano @salaxander Hey raised their PR which fixes the issue #355

PR #389, opened 2025-04-09
Improve Config API reference generator for resource types

tengqm:
The `types.go` file for some APIs may not have the anticipated `// +k8s:deepcopy-gen:...` line for any types. We have to allow the generator to identify the "exposed" resource types. For example, the `Summary` type from the kubernetes stats API does not have this comment line.


neolit123 (commented):
/lgtm
/hold

please drop hold when needed.

PR #397, opened 2025-06-28
Fix: Remove duplicated short command descriptions in generated kubeadm CLI docs.

Mujib-Ahasan:
This PR is for to remove duplicated command descriptions above the Synopsis section for both phase - `kubeadm upgrade apply phase` and `kubeadm upgrade node phase` in the kubeadm upgrade documentation. It solves the same issue for few others kubeadm docs also.

Main Page: https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-upgrade-phase/

Closes: kubernetes/website#51172

### DIFF
--- a/gen-compdocs/generators/doc.go
+++ b/gen-compdocs/generators/doc.go
@@ -155,9 +155,13 @@ func GenReference(cmd *cobra.Command, w io.Writer, linkHandler func(string) stri
 		// Print the "generated" warning
 		fmt.Fprintf(w, "%s\n\n", generated_warning)
 
-		if _, err := fmt.Fprintf(w, "%s\n\n", short); err != nil {
-			return err
-		}
+		trimmedShort := strings.TrimSpace(strings.TrimSuffix(short, "."))
+        trimmedLong := strings.TrimSpace(strings.TrimSuffix(long, "."))
+		if !strings.HasPrefix(trimmedLong, trimmedShort) {
+	        if _, err := fmt.Fprintf(w, "%s\n\n", short); err != nil {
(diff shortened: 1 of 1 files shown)


Mujib-Ahasan:
Hey @tengqm, Whenever you get a chance, could you please take a look at this PR. I'd appreciate any feedback or suggestions.

gen-compdocs/generators/doc.go
```diff
@@ -155,9 +155,13 @@ func GenReference(cmd *cobra.Command, w io.Writer, linkHandler func(string) stri
 		// Print the "generated" warning
 		fmt.Fprintf(w, "%s\n\n", generated_warning)
 
-		if _, err := fmt.Fprintf(w, "%s\n\n", short); err != nil {
-			return err
-		}
+		trimmedShort := strings.TrimSpace(strings.TrimSuffix(short, "."))
+        trimmedLong := strings.TrimSpace(strings.TrimSuffix(long, "."))
```
tengqm, line 159: `TrimSpace` makes sense to me, why `TrimSuffix` ?

Mujib-Ahasan: For a case, if my "short" description has full stop (".") at the end then comparing them both would not show expected behavior in `HasPrefix` method. I used `TrimSuffix` for "long" description also just to bypass any unwanted cases but I think omitting it for "long" would not disturb the result.

tengqm: Fine. Sounds like a limitation of the `HasPrefix` method.

PR #399, opened 2025-09-04
Update component reference generator for v1.34

tengqm:
This PR updates the component reference generator for v1.34.


gen-compdocs/comps/gen_kube_docs.go
```diff
@@ -55,9 +55,9 @@ func GenerateFiles(path, module string) error {
 		scheduler := schapp.NewSchedulerCommand()
 		return generators.GenMarkdownTree(scheduler, outDir, true)
 
-	case "kubelet":
-		kubelet := kubeletapp.NewKubeletCommand()
-		return generators.GenMarkdownTree(kubelet, outDir, true)
+	//	case "kubelet":
+	//	kubelet := kubeletapp.NewKubeletCommand()
```
soltysh, line 59: This should solve the compilation problem you're seeing:
```golang
kubelet := kubeletapp.NewKubeletCommand(context.TODO())
```
and let's get this back. After that is fixed I'll tag.

tengqm: @soltysh The generator has been updated. However, as you can see, the generated kubelet.md file doesn't contain any useful information. I am not quite sure if this is related to the change mtaufen merged several years ago (https://github.com/kubernetes/kubernetes/blame/master/cmd/kubelet/app/server.go#L172). After those refactoring, we had never been able to generate meaningful reference for kubelet. Do you have any ideas to restore this capability?

gen-compdocs/generators/doc.go
```diff
@@ -391,6 +392,9 @@ func flagUsages(f *pflag.FlagSet) string {
 		line += "</td>\n</tr>\n<tr>\n<td></td><td style=\"line-height: 130%; word-wrap: break-word;\">"
 
 		// process markdown in usage, force wrap for "\n"
+		if len(flag.Deprecated) > 0 {
```
soltysh, line 395: I see you've decided to also print deprecated - nice :)

soltysh (approved):
/lgtm

gen-compdocs/comps/gen_kube_docs.go
soltysh: That's my hope as well :sweat_smile:

PR #401, opened 2025-09-04
Adapt config for v1.34 API reference generator

tengqm:
This PR adapts the configuration file for API reference generator v1.34.

PR #400, opened 2025-09-04
Update resource API generator for v1.34

tengqm:
This PR updates the resource API reference generator for v1.34.

PR #406, opened 2025-09-15
Update generators with k8s 1.34

soltysh:
This PR does the following:
1. Updates Makefiles, fixing var names, and adds more errors so it's easier to run `make`
~2. Bumps k8s to 1.34, where needed.~
3. Runs all generators.

/assign @tengqm

### DIFF
--- a/gen-apidocs/config/config.yaml
+++ b/gen-apidocs/config/config.yaml
@@ -114,7 +114,7 @@ resource_categories:
       version: v1
       group: storage
     - name: VolumeAttributesClass
-      version: v1beta1
+      version: v1
       group: storage
   - name: "Metadata APIs"
     include: "meta"
@@ -129,7 +129,7 @@ resource_categories:
       version: v1
       group: apiextensions
     - name: DeviceClass
-      version: v1beta2
+      version: v1
       group: resource
     - name: DeviceTaintRule
       version: v1alpha3
@@ -156,13 +156,13 @@ resource_categories:
       version: v1
       group: scheduling
     - name: ResourceClaim
-      version: v1beta2
+      version: v1
       group: resource
     - name: ResourceClaimTemplate
-      version: v1beta2
+      version: v1
       group: resource
     - name: ResourceSlice
-      version: v1beta2
+      version: v1
       group: resource
     - name: ValidatingAdmissionPolicy
       version: v1
@@ -219,6 +219,10 @@ resource_categories:
       version: v1
       group: core
       description_note: "These are assigned to <a href=\"#pod-v1-core\">Pods</a> using <a href=\"#persistentvolumeclaim-v1-core\">PersistentVolumeClaims</a>."
+    - name: PodCertificateRequest
+      version: v1alpha1
+      group: certificates
+
     - name: PriorityLevelConfiguration
       version: v1
       group: flowcontrol
@@ -268,10 +272,10 @@ resource_categories:
       version: v1
       group: networking
     - name: MutatingAdmissionPolicy
-      version: v1alpha1
+      version: v1beta1
       group: admissionregistration
     - name: MutatingAdmissionPolicyBinding
-      version: v1alpha1
+      version: v1beta1
       group: admissionregistration
 
 operation_categories:
--- a/gen-kubectldocs/generators/v1_34/static_includes/_app_management.md
+++ b/gen-kubectldocs/generators/v1_34/static_includes/_app_management.md
@@ -1,0 +1,4 @@
+# <strong>APP MANAGEMENT</strong>
+
+This section contains commands for creating, updating, deleting, and
+viewing your workloads in a Kubernetes cluster.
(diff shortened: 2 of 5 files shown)


soltysh:
I've also opened https://github.com/kubernetes/website/pull/52354 to update the missing artifacts on the main k8s website.

gen-apidocs/config/config.yaml
```diff
@@ -144,6 +155,24 @@ resource_categories:
     - name: PriorityClass
       version: v1
       group: scheduling
+    - name: ResourceClaim
+      version: v1beta2
```
lmktfy, line 159: Shouldn't this be v1?

gen-apidocs/config/config.yaml
```diff
@@ -144,6 +155,24 @@ resource_categories:
...
+      group: resource
+    - name: ResourceClaimTemplate
+      version: v1beta2
+      group: resource
+    - name: ResourceSlice
+      version: v1beta2
```
lmktfy, line 165: Shouldn't this be v1?

gen-apidocs/config/config.yaml
```diff
@@ -102,24 +104,36 @@ resource_categories:
...
+      version: v1beta1
+      group: certificates
     - name: ControllerRevision
       version: v1
       group: apps
     - name: CustomResourceDefinition
       version: v1
       group: apiextensions
+    - name: DeviceClass
+      version: v1beta2
```
lmktfy, line 132: Shouldn't this be v1?

gen-apidocs/config/v1_34/config.yaml
```diff
@@ -0,0 +1,391 @@
...
+      group: authentication
+    - name: NetworkPolicy
+      version: v1
+      group: networking
+    - name: MutatingAdmissionPolicy
+      version: v1alpha1
+      group: admissionregistration
+    - name: MutatingAdmissionPolicyBinding
+      version: v1alpha1
+      group: admissionregistration
```
lmktfy, line 275: These actually belong within Policy.

gen-compdocs/comps/gen_kube_docs.go
```diff
@@ -56,7 +57,7 @@ func GenerateFiles(path, module string) error {
 		return generators.GenMarkdownTree(scheduler, outDir, true)
 
 	case "kubelet":
-		kubelet := kubeletapp.NewKubeletCommand()
+		kubelet := kubeletapp.NewKubeletCommand(context.TODO())
```
lmktfy, line 60: Is this right?

lmktfy (commented):
To be honest, I don't know how this code works. I can only point out the places where the PR doesn't look correct; I'm not in an ideal position to comment on the good stuff.

gen-apidocs/config/config.yaml
soltysh: This and other comments are probably b/c I missed something, which @tengqm fixed in his PRs.

soltysh:
> Please check existing PRs that adapt the generators to v1.34 releases.
> Although these tools live in the same repo, they are maintained separately.
> Hope this makes sense.

Yeah, that makes total sense.

> #399, #400, #401, #402

@tengqm I've tagged the latter 3 PRs, the first has a small nit to be fixed so that you can continue generating completion docs for kubelet. 

Let's hold with this PR, until all the other merge, and I'll check if it's still needed afterwards. Probably some of the `Makefile` will be the only ones that will be left.

/hold

soltysh:
This should address all the comments from @lmktfy, especially around the API versions, not sure why previously it was picking the wrong ones, probably something we'll need to look.

@tengqm ptal, there are still some minor, generated changes that I think are worth including. Once this merges I'd like to refresh https://github.com/kubernetes/website/pull/52354 with that data.

Makefile
```diff
@@ -113,22 +120,13 @@ copyapi: api
...
-	cp $(CONFIGSRC)/apiserver-webhookadmission.v1.md $(CONFGDST)
-	cp $(CONFIGSRC)/client-authentication.v1beta1.md $(CONFGDST)
-	cp $(CONFIGSRC)/kube-proxy-config.v1alpha1.md $(CONFGDST)
-	cp $(CONFIGSRC)/kube-scheduler-config.v1beta1.md $(CONFGDST)
-	cp $(CONFIGSRC)/kube-scheduler-policy-config.v1.md $(CONFGDST)
-	cp $(CONFIGSRC)/kubelet-config.v1beta1.md $(CONFGDST)
-	cp $(CONFIGSRC)/kuberc.v1alpha1.md $(CONFGDST)
-	cp $(CONFIGSRC)/kuberc.v1beta1.md $(CONFGDST)
+copyconfigapi: configapi
+	cp $(CONFIGSRC)/*.md $(CONFIGDST)/
```
tengqm, line 132: The collection of so called "config APIs" is larger than what we publish on k8s/website.
The reason is that not all markdowns are referenced from the website pages.
It is fine to copy them all over, but a tighter control might be better.
We can leave that to a future PR.

soltysh: My goal was to ensure that new ones are copied over, and when creating a PR against k/website that's where you can decide what should and should not be included.

PR #407, opened 2025-09-17
refactor: simplify code logic and loops in `util.go`

lavishpal:
This PR address the several changes in `util.go` file:

1. Replaced multiple iterations over `definitions.All `with a single loop.
2. Simplified conditions to reduce nested `if `statements.


lmktfy:
I don't know how to run this to compare the old and the new code. I can put some time aside to learn that process, but I already have plenty on … maybe later October?

PR #408, opened 2025-09-24
Make `k8s.io/kubernetes` repo source path detection flexible

lavishpal:
This PR removes the need of renaming the kubernetes/kubernetes repo to by:

Updating the Makefile to auto-detect the Kubernetes source in common locations:

```
../kubernetes

../../kubernetes

../../../k8s.io/kubernetes

../../../../k8s.io/kubernetes

```

Update `.travis.yml `to use a configurable `K8S_SRC_PATH` that gets passed to the Makefile as `K8S_ROOT.`
Keeping backward compatibility for setups where the repo is already cloned as `k8s.io/kubernetes.`

Now Contributors don’t need to rename their Kubernetes repo to` k8s.io/kubernetes.` The build and CI now work with common repo clone paths and also the existing workflows remain supported.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ install:
 - mkdir -p ${BUILD_PATH}/src/k8s.io
 - wget https://github.com/kubernetes/kubernetes/archive/v${K8S_RELEASE}.tar.gz -O ${BUILD_PATH}/src/k8s.io/kubernetes-src.tar.gz
 - pushd ${BUILD_PATH}/src/k8s.io && tar xzf kubernetes-src.tar.gz && mv kubernetes-${K8S_RELEASE} kubernetes && popd
+- # Export K8S_ROOT for API docs generation (updateapispec target)
+- export K8S_ROOT=${BUILD_PATH}/src/k8s.io/kubernetes
 - pushd ${BUILD_PATH}/src/k8s.io/kubernetes && make generated_files && popd
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,6 @@ install:
 - mkdir -p ${BUILD_PATH}/src/k8s.io
 - wget https://github.com/kubernetes/kubernetes/archive/v${K8S_RELEASE}.tar.gz -O ${BUILD_PATH}/src/k8s.io/kubernetes-src.tar.gz
 - pushd ${BUILD_PATH}/src/k8s.io && tar xzf kubernetes-src.tar.gz && mv kubernetes-${K8S_RELEASE} kubernetes && popd
-- # Export K8S_ROOT for API docs generation (updateapispec target)
-- export K8S_ROOT=${BUILD_PATH}/src/k8s.io/kubernetes
 - pushd ${BUILD_PATH}/src/k8s.io/kubernetes && make generated_files && popd
 
 script:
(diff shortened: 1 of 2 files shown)


lavishpal:
cc: @tengqm , @katcosgrove , @salaxander , @reylejano

tengqm:
I don't think this path detection logic helps simplify the reference generation process.
I failed to understand the reason why we need to search these paths.
Instead, a more straightforward way is to make sure K8S_ROOT environment variable is set and it contains a valid branch/tag of the Kubernetes release.

lavishpal:
The main issue I was trying to address is that currently the build fails unless the Kubernetes repo is renamed to `k8s.io/kubernetes`. Many contributors clone the repo in different locations (e.g., `~/kubernetes `or inside `github.com/kubernetes/kubernetes`) and then run into errors because of the hardcoded path requirement and my attempt to reduce that by resolving the Kubernetes source if `K8S_ROOT` is not explicitly set.

natalisucks:
> I think the intent is to remove a step in generating the reference docs

This was also my read of the intent. @lavishpal do you have other feedback regarding contributors cloning the repo and running into problems when generating the reference docs? Some examples of errors might be useful here so folks understand intent

tengqm:
> > > I think the intent is to remove a step in generating the reference docs
> > 
> > 
> > This was also my read of the intent. @lavishpal do you have other feedback regarding contributors cloning the repo and running into problems when generating the reference docs? Some examples of errors might be useful here so folks understand intent
> 
> Yes @natalisucks , the intent was to remove the extra step where contributors had to rename their Kubernetes repo to `k8s.io/kubernetes`. But this issue typically shows up when the repo is cloned as `~/kubernetes` or `github.com/kubernetes/kubernetes` instead of under `k8s.io/kubernetes`, which then causes the reference docs generation to fail.

This is how the Kubernetes project works. All sources import kubernetes and its subprojects from the 'k8s.io' domain. For example, 'k8s.io/api', 'k8s.io/kubelet', 'k8s.io/website'. It is not what you see from github.com.

The real fix/improvement is to make sure the 'kubernetes' project is placed under `/<somewhere>/k8s.io/kubernetes` so that we won't encounter many surprises.

A common/desired layout for the directory is something like this:

```
<work dir> 
   |-- k8s.io
   |     |-- kubernetes
   |     |-- website
   |     `-- <other projects>
   |-- github.com
   `-- <other domains>
```

This can be easily achieved using the 'git clone' commands with the help of some bash commands. With the source repositories properly placed, we don't (and we should not) search the world for another project. In other words, I don't see a reason why kubernetes lives in a relative path such as `../../../kubernetes`.

tengqm:
As for the change to the travis file, it is not an improvement in my view.
We have control on where the repos are cloned, and this is especially true for the travis environment. It is not exposed to a huge team for customization. It is not something we change every month.

lavishpal:
Intent was to address the issues contributors face locally when their repo isn’t under `k8s.io/kubernetes`. I agree that travis file is controlled and stable environment, so dynamic path detection there isn’t needed. So, I’ll update the PR and revert the changes in the `travis.yml`.
Any thoughts on this @tengqm , @natalisucks , @reylejano , @katcosgrove

salaxander:
So ultimately I don't like forcing users to rename a directory, but I also agree with @tengqm that searching directories for a k8s clone is not ideal. I think it would be ok to require users to set an environment variable as long as that's very clearly expressed in error logging, as that's pretty standard practice for CLI tooling

PR #414, opened 2025-10-22
[gen-apidocs] fix the error wrapping and input validation 

lavishpal:
This PR fixes

1. Improve the error wrapping and remove the dead code in the `definition.go` file.
2. fix the input validation and grouped both regex definitions into a single `var (...)` block in the `operation.go file`.

### DIFF
--- a/gen-apidocs/generators/api/definition.go
+++ b/gen-apidocs/generators/api/definition.go
@@ -171,9 +171,22 @@ func (s *Definitions) getReferences(d *Definition) []*Definition {
 
 	/*
 
-				// TODO: Support nested additionalProperties
-		        // Currently only handles top-level additionalProperties. Nested structures require
-				// more complex handling and are not yet supported.
+		    The following logic is an attempt to probe nested additionalProperties, but it is not working
+			yet because the additionalProperties can be deeply nested rather than being a top-layer property.
+
+			if d.schema.AdditionalProperties != nil {
+				as := d.schema.AdditionalProperties.Schema
+				if as != nil && IsComplex(*as) {
+					// Look up the definition for the referenced definitions
+					if schema, ok := s.GetForSchema(*as); ok {
+						fmt.Printf("*** schema: %s\n", schema)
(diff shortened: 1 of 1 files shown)


gen-apidocs/generators/api/definition.go
```diff
@@ -170,22 +170,10 @@ func (s *Definitions) getReferences(d *Definition) []*Definition {
...
-				} else {
-					g, v, k := GetDefinitionVersionKind(*as)
-					fmt.Printf("Could not locate referenced property of %s: %s (%s/%s).\n", d.Name, g, k, v)
-				}
-			}
-		}
+
+				// TODO: Support nested additionalProperties
+		        // Currently only handles top-level additionalProperties. Nested structures require
+				// more complex handling and are not yet supported.
```
tengqm: This change is not needed

lavishpal: Can you elaborate on this a little bit?
I have added this TODO because i removed the dead code in this file.

tengqm: The code, although commented out, is still a good reference when we want to improve the logic in the future.
What do we gain from removing these lines?

tengqm: Do you mean code that are commented out is a cognitive overhead?
Why is that?
Why the revised comment about git history is a heavier cognitive overhead for me?

tengqm: Since you are already on to this, why don't you complete this logic and make it work?

lavishpal: I’m absolutely open to implementing this properly.I just didn't want to mix cleanup and feature implementation in the same PR.

If you prefer, I can create a separate  follow up PR to complete this logic .

lavishpal: Thanks for the clarification @reylejano , i will revert the changes in this pr and implement the commented code logic in the follow up pr.

PR #432, opened 2026-04-04
gen-apidocs: default to auto-detect and trim v1.35 group mappings

Caesarsage:
Follow up PR reduces manual API config maintenance for v1_35 by relying on the already-merged --auto-detect behavior during API generation.

This is a follow-up PR for https://github.com/kubernetes-sigs/reference-docs/pull/431 to safely removes redundant manual entries for v1_35 and keeps only curated/special mappings.

## Validation
Ran API generation and compared output against auto-detect baseline.
index.html diff showed timestamp-only change (Generated at), no content drift.

This indicates the config cleanup is behaviorally safe for the auto-detect generation path.

image

## Notes
- This PR intentionally does not change operation_group_map.
- Scope is limited to v1_35 config cleanup + Makefile default behavio


tengqm (commented):
/lgtm
/approve

Nice.

PR #433, opened 2026-04-05
Auto add new resources

Caesarsage:
Extends `--auto-detect` to auto-generate resource categories from API groups in swagger.json and auto-add new resources to the ToC. 

## What it does

- **Group-based categories**: Replaces curated categories (Workloads, Service, etc.)
  with API-group-based categories (Apps, Batch, Core, Networking, etc.) when
  `--auto-detect` is enabled.
- **Auto-adds new resources**: Any top-level resource in swagger.json is included
  automatically — no manual config.yaml entry needed.
- **Preserves annotations**: `description_warning` and `description_note` from
  config.yaml are merged into the auto-generated categories.
- **Preserves non-top-level annotated resources**: Editorail only entry like `Container` are retained.
- **config.yaml cleanup**: Removes all resource entries except the 6 with annotations.

## Test

With --auto-detect (group-based categories)

go run main.go --kubernetes-release=1.35 --work-dir=. --auto-detect
open build/index.html

image


gen-apidocs/config/v1_35/config.yaml
```diff
@@ -1,261 +1,39 @@
...
-    - name: DaemonSet
-      version: v1
-      group: apps
-    - name: Deployment
-      version: v1
-      group: apps
-    - name: Job
-      version: v1
-      group: batch
+      description_warning: "Containers are only ever created within the context of a <a href=\"#pod-v1-core\">Pod</a>. This is usually done using a Controller. See Controllers: <a href=\"#deployment-v1-apps\">Deployment</a>, <a href=\"#job-v1-batch\">Job</a>, or <a href=\"#statefulset-v1-apps\">StatefulSet</a>"
```
tengqm: Actually, these warnings can be safely dropped. I was hoping that this kind of information would be useful to users. However, it turned out that no one cares.

tengqm (commented):
Overall looks good.
I'd suggest we drop those "description_xxx" fields in preference of the code maintainability.
WDYT?

Caesarsage:
> Overall looks good. I'd suggest we drop those "description_xxx" fields in preference of the code maintainability. WDYT?

Makes sense, I'll drop the description_xxx entries from config.yaml. But I suggest we still keep the fields in the Go structs and merge logic, so they remain available if ever needed in the future for any override. Will not have any maintenance cost since there are no entries to process.

tengqm:
> > Overall looks good. I'd suggest we drop those "description_xxx" fields in preference of the code maintainability. WDYT?
> 
> Makes sense, I'll drop the description_xxx entries from config.yaml. But I suggest we still keep the fields in the Go structs and merge logic, so they remain available if ever needed in the future for any override. Will not have any maintenance cost since there are no entries to process.

This is a solid step. Thank you.

By the way, can you try inspect the generated HTML and see the CSS/JS files included? The paths for those files were chosen based on the layout of the K8s website. Maybe you can setup a local web server to mimic that directory layout so that you can get an exact look and feel on your local machine.

tengqm (commented):
/lgtm
/approve

Caesarsage:
> > > Overall looks good. I'd suggest we drop those "description_xxx" fields in preference of the code maintainability. WDYT?
> > 
> > 
> > Makes sense, I'll drop the description_xxx entries from config.yaml. But I suggest we still keep the fields in the Go structs and merge logic, so they remain available if ever needed in the future for any override. Will not have any maintenance cost since there are no entries to process.
> 
> This is a solid step. Thank you.
> 
> By the way, can you try inspect the generated HTML and see the CSS/JS files included? The paths for those files were chosen based on the layout of the K8s website. Maybe you can setup a local web server to mimic that directory layout so that you can get an exact look and feel on your local machine.

Look Good

image

Issue #434, opened 2026-04-18
Merge gen-resourcesdocs into gen-apidocs


Caesarsage:
`gen-apidocs` and `gen-resourcesdocs` share the same swagger input but produce
different outputs (single-page HTML vs. Hugo markdown per resource). Duplicated
ingestion, duplicated config, two places for `toc.yaml` drift on version bumps.

Proposal: 

extend `gen-apidocs` with a markdown backend matching
`gen-resourcesdocs/templates/chapter.tmpl`, then deprecate `gen-resourcesdocs`
once output parity is verified.

lmktfy:
The `chapter` part isn't especially important IMO. We could group the generated Markdown files in directories named after their API group (I would).

If it's important to have different groupings, as https://kubernetes.io/docs/reference/kubernetes-api/ has today, then the data to drive that can live inside k/website and Hugo can take care of building a list / index page at site build time. That's the kind of work that Hugo is good at.

Caesarsage:
Thanks, I agree the chapter is not needed, I'm only using the template. I'll update #434's description to reflect that.

On the group-by-API-group point: we're aligned. PR 1 (#435) already groups by API group via auto-detect (building on the categorization work in #431, which replaced manual group curation). So the remaining mismatch is field rendering with Hugo shortcodes, per #440.

I'd like to finish inline nested type recursion first, that's the next PR in the stack. It will settle how nested fields are represented in the data model, and then we can design shortcodes that mirror that structure rather than re-doing the model twice

PR #435, opened 2026-04-19
extend gen-apidocs with a markdown backend

Caesarsage:
Implements the `MarkdownWriter` backend in `gen-apidocs`, producing
Hugo-compatible markdown output for Kubernetes API reference pages.

Output shape targets parity with `gen-resourcesdocs/templates/chapter.tmpl`,
with one deliberate UX change per feedback: resource attributes render
as **HTML tables** instead of bullet lists. Template lives at
`gen-apidocs/generators/templates/resource.tmpl` for easier maintenance.

### What's NOT in this PR

Deferred to follow-up PRs so reviews stay bounded:

- cross-type `{{< ref >}}` link resolution (two-pass render)
- inline nested type recursion (dot-notation rows for inlined types)
- deprecate `gen-resourcesdocs` (pending output parity verification)

part of #434


tengqm:
so far so good 

While I'm supportive to generating markdown format, per-resource and/or per-group files, I am not sure that markdown is always the best for all scenarios. For example, when listing resource attributes, I'd still prefer embedded HTML tables.

There are many other aspects where we can improve. Maybe you can give `docker run -p 8000:8000 tengqm/kab` a try?

Caesarsage:
> so far so good
> 
> While I'm supportive to generating markdown format, per-resource and/or per-group files, I am not sure that markdown is always the best for all scenarios. For example, when listing resource attributes, I'd still prefer embedded HTML tables.
> 
> There are many other aspects where we can improve. Maybe you can give `docker run -p 8000:8000 tengqm/kab` a try?

Thanks, I will check this out.

I was trying to use similar pattern as the genresources and since the k/website is on markdown

Caesarsage:
using HTML tables will change the UX of the current k/website ref generator. (i think it will even make it better). Do we want to go with that @tengqm ?

Caesarsage:
IMO, I think table view is better

screencapture-localhost-1313-docs-reference-kubernetes-api-workload-resources-deployment-v1-preview-tables-2026-04-21-13_10_39

lmktfy:
I wonder if we're actually OK to emit Hugo shortcodes. Coupling to Hugo feels OK.

Within the shortcode, the contents can be HTML (tables), or even YAML, or something like:

```markdown
{{< api-reference-toplevel >}}
{{< api-reference-description >}}
---
# Hugo-style front matter (YAML)
kind: Deployment
apiGroup: "" # core
apiVersion: "v1"
fieldName: "spec" # this section is for .spec in a Deployment
---
DeploymentSpec is the specification of the desired behavior of the Deployment.
{{< /api-reference-description >}}
{{< api-reference-subfield >}}
---
# Hugo-style front matter (YAML)
apiKind: Deployment
topLevelFieldName: "spec"
childFieldName: "selector"
fieldType: LabelSelector
fieldRequired: true
---
Label selector for pods. Existing ReplicaSets whose pods are selected by this will be the ones affected by this deployment. It must match the pod template's labels.
{{< /api-reference-subfield >}}
{{< api-reference-subfield >}}
---
# Hugo-style front matter (YAML)
apiKind: Deployment
topLevelFieldName: "spec"
childFieldName: "template"
fieldType: PodTemplateSpec
fieldRequired: true
---
Template describes the pods that will be created. The only allowed `template.spec.restartPolicy` value is "Always".
{{< /api-reference-subfield >}}
…
{{< /api-reference-toplevel >}}
```

Caesarsage:
Good direction, separating data from presentation is a good architectural improvement.

The goal of this PR is narrower though: deprecate gen-resourcesdocs by having gen-apidocs produce equivalent output with HTML tables instead of bullet lists. 

The shortcode-DSL is a real improvement but it's a cross-repo design RFC, shortcodes have to exist in k/website before gen-apidocs can emit them. Happy to pursue as a follow-up once this automation lands.

tengqm:
Agreed. A good idea worth experimenting. Maybe we can file a feature request issue so that we can follow up later.

Caesarsage:
Sounds good, I'll file a feature request issue capturing the shortcode-DSL direction so we can pick it up after consolidation lands

tengqm:
I still believe HTML table view is a lot better.
The reasons people wanted a "markdown" version of API reference are:

- the single-page version API reference takes time to load
- the single-page version too long to find resource definitions
- the single-page version was using a separate full-page view, which deviates from the website page organization

However, organizing field definitions in tables is really neat, in my view. I heard no objections to this.

PR #442, opened 2026-05-03
regenerate swagger from k/k hack/update-openapi-spec.sh (OpenAPIEnums…

Caesarsage:
…=true) to update missing enum


tengqm (commented):
/lgtm
/approve

PR #444, opened 2026-05-03
Update/api docs openapi enums

Caesarsage:
Regenerates the v1.36 OpenAPI swagger with enum metadata and updates markdown generation so possible enum descriptions render cleanly in generated API reference tables.

Also separates `make api` and `make apimd` outputs into gen-apidocs/build/html and gen-apidocs/build/markdown, so running one target no longer deletes the other backend’s output.


Caesarsage:
With this, next will be to add a depreciation not on the resources-gen and update k/website for new look

PR #447, opened 2026-05-07
fix(gen-apidocs): bucket ByKind by group+kind to avoid cross-group collisions

### LINKED ISSUE #446: gen-apidocs: ByKind keyed by name only causes cross-group IsOldVersion false positives
## Problem

`Definitions.ByKind` in [generators/api/definition.go](https://github.com/kubernetes-sigs/reference-docs/blob/master/gen-apidocs/generators/api/definition.go) is keyed by kind name only:

```go
for _, d:= range s.All {
    s.ByKind[d.Name] = append(s.ByKind[d.Name], d)
}
```

When two unrelated API groups define the same kind name, they share a bucket. The next loop sorts the bucket and flags everything past index 0 as IsOldVersion, which is wrong when the entries belong to different groups.

## Example HTML backend

Take `TypedLocalObjectReference` as a concrete case. It exists as two unrelated kinds:
 
- `core/v1.TypedLocalObjectReference`
- `scheduling.k8s.io/v1alpha2.TypedLocalObjectReference`

They share a name but are independent types. With name-only `ByKind` keying, the
generator treats them as versions of each other.

**Master (buggy):**

Image

> The `scheduling.k8s.io/v1alpha2` page shows an "Other API versions of this object
> exist: v1" alert linking to `core/v1.TypedLocalObjectReference`: a misleading
> claim, since the two are independent kinds.

**With the fix:**

Image

> The misleading alert is gone. Each kind is treated as its own bucket.

## Proposed fix

Key `ByKind` by `(group, kind)` instead of `kind` alone:

```go
type byKindKey struct {
    Group string
    Kind  string
}

type Definitions struct {
    ByKind map[byKindKey]SortDefinitionsByVersion
    ...
}
```

`FindNewestVersion(group, kind)` already takes a group; manual filtering becomes unnecessary. Same v1.36 spec, after the fix: zero cross-group false positives.

PR follows.

Caesarsage:
Key `ByKind` by `(group, kind)` instead of `kind` alone:

Closes: #446

### DIFF
--- a/gen-apidocs/generators/api/definition.go
+++ b/gen-apidocs/generators/api/definition.go
@@ -49,7 +49,7 @@ var _INLINE_DEFINITIONS = []inlineDefinition{
 func NewDefinitions(config *Config, specs []*loads.Document) (*Definitions, error) {
 	s := &Definitions{
 		All:           map[string]*Definition{},
-		ByKind:        map[string]SortDefinitionsByVersion{},
+		ByKind:        map[byKindKey]SortDefinitionsByVersion{},
 		GroupVersions: map[string]ApiVersions{},
 	}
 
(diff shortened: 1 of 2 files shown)


tengqm (commented):
/lgtm
/approve

Issue #448, opened 2026-05-07
[Umbrella] Release tooling consolidation: Makefile cleanup, deprecate gen-kubectldocs, add copy targets, switch website to apimd


Caesarsage:
## Goal

Tidy up the release-team experience after the gen-apidocs markdown backend
landed. The Makefile has accumulated drift, gen-kubectldocs is functionally
replaced by gen-compdocs but still present, and several copy targets are
missing. Each of these is small in isolation but coordinated together they
make the per-release workflow significantly cleaner.
## Sub-tasks

1. [X] **Makefile cleanup** PR (https://github.com/kubernetes-sigs/reference-docs/pull/449)
   - Move `K8S_ROOT` / `K8S_RELEASE` / `K8S_WEBROOT` checks into the recipes
     that need them (currently fail every `make` invocation)
   - Drop the duplicate `K8S_ROOT` check
   - Drop the unused `CURDIR=$(shell pwd)` line in `updateapispec`
   - Add `copyapimd` target (symmetric with `copyapi`) for the markdown
     output → `content/en/docs/reference/kubernetes-api/`
   - Refresh the `all` target's printed list of supported targets

2. [X] **Deprecate gen-kubectldocs** (https://github.com/kubernetes-sigs/reference-docs/pull/450)
   - Same pattern as the gen-resourcesdocs deprecation in
     [#445](https://github.com/kubernetes-sigs/reference-docs/pull/445):
     add a deprecation banner to `gen-kubectldocs/README.md`
   - Add a `WARNING:` echo to `cli`, `copycli`, `cleancli` targets pointing
     users to gen-compdocs

3. [X] **Add `copycomp` family of targets** (https://github.com/kubernetes-sigs/reference-docs/pull/452)
   - `copycomp-core`: kube-apiserver, kube-controller-manager, kube-scheduler,
     kube-proxy → `content/en/docs/reference/command-line-tools-reference/`
   - `copycomp-kubeadm`: kubeadm tree → `content/en/docs/reference/setup-tools/kubeadm/generated/`
   - `copycomp-kubectl`: kubectl tree → `content/en/docs/reference/kubectl/generated/`
   - Mirrors @tengqm's 3-PR component handoff flow.

4. [X] **Fix cross-group `ByKind` collision** PR (https://github.com/kubernetes-sigs/reference-docs/pull/447)
   - Tracked at #446 with concrete v1.36 evidence (cross-group "Other API
     versions" leaks in HTML output)
[...]

PR #449, opened 2026-05-07
chore(makefile): scope env-var checks per target, add copyapimd, refresh help

Caesarsage:
Makefile cleanup
- Move K8S_ROOT / K8S_RELEASE / K8S_WEBROOT checks into the recipes
that need them (currently fail every make invocation)
- Drop the duplicate K8S_ROOT check
- Drop the unused CURDIR=$(shell pwd) line in updateapispec
- Add copyapimd target (symmetric with copyapi) for the markdown
output → content/en/docs/reference/kubernetes-api/
- Refresh the all target's printed list of supported targets

related to https://github.com/kubernetes-sigs/reference-docs/issues/448

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,36 +19,33 @@
 
 
 WEBROOT=${K8S_WEBROOT}
-
-# Validate K8S_ROOT is set and points to valid Kubernetes source
-ifeq ($(K8S_ROOT),)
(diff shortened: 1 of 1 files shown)


tengqm (commented):
Thanks for cleaning up the mess accumulated so far.
/lgtm
/approve

PR #451, opened 2026-05-13
reduce gen-apidocs markdown output from 577 to 157 pages

Caesarsage:
The markdown backend's output made k/website's `hugo server` dev mode
unusable (initial build minutes, home page sometimes wouldn't render).

## Approach
Four commits:
1. Skip old-version pages. (similar to old resource-gen approach)
2. Hide `definitions/` from Docsy sidebar nav.
3. Refcount-based inlining of non-resource types into their closest
   top-level parent with BFS. Two carve-outs: `io.k8s.apimachinery.*` types and a
   4-name core/v1 allowlist stay standalone.
4. Drop duplicate H1 on category index pages.

## Result
| | Before | After | Upstream baseline |
|---|---|---|---|
| Pages | 577 | **157** | 101 |
| Dev build | unusable | 120s | 42s |

## Companion PR
https://github.com/kubernetes/website/pull/55758

related #448

### DIFF
--- a/gen-apidocs/generators/markdown.go
+++ b/gen-apidocs/generators/markdown.go
@@ -267,8 +267,6 @@ func (m *MarkdownWriter) WriteResourceCategory(name, file string) error {
 
 	if body := readOptionalSection(file + ".md"); body != "" {
 		fmt.Fprintln(f, body)
-	} else {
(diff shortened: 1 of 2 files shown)


tengqm (commented):
/lgtm
/approve

PR #455, opened 2026-05-28
change '<','>' only if it is treated as a html tag

ingyeoking13:
# Why

[Generated docs kubectl explain has typo with describing JsonPath Identifier](https://github.com/kubernetes/website/issues/54776)

# How

change
from: replace '<', '>' to '&lt;, &gt;' always 
to: replace  '<', '>' to '&lt;' &gt' only if it is treated as html tag (unexpected in long description)

# affected docs(generated docs) after change

1. kubectl explain

```
after <         <type>.<fieldName>[.<fieldName>]
---
before >         &lt;type&gt;.&lt;fieldName&gt;[.&lt;fieldName&gt;]
```

2. kubectl diff

```
after <  Exit status: 0 No differences were found. 1 Differences were found. >1 Kubectl or diff failed with an error.
---
before >  Exit status: 0 No differences were found. 1 Differences were found. &gt;1 Kubectl or diff failed with an error.
```

3. kubectl completion

```
after <  Note for zsh users: [1] zsh completions are only supported in versions of zsh >= 5.2.
---
before >  Note for zsh users: [1] zsh completions are only supported in versions of zsh &gt;= 5.2.
```

# Test

test with [kubernetes/website](https://github.com/kubernetes/website) and [kubernetes/kubernetes](https://github.com/kubernetes/kubernetes)

# Comment

This is my first PR in k8s Project. and If I missed something, just let me know.  
Thank you!

### DIFF
--- a/gen-compdocs/generators/doc.go
+++ b/gen-compdocs/generators/doc.go
@@ -528,28 +528,8 @@ func escapeAngleBracket(long string) string {
 			return ast.WalkContinue, nil
 		}
 
-		// when pair of '<', '>' characters found in the text and it is treated as a RawHTML node,
-		// then escape the text in RawHTML nodes.
-		t, ok := n.(*ast.RawHTML)
-		if !ok {
-			return ast.WalkContinue, nil
-		}
-
-		segs := t.Segments
-
-		for i := 0; i < segs.Len(); i++ {
-			seg := segs.At(i)
-			changing := seg.Value(source)
-			changing = bytes.ReplaceAll(changing, []byte("<"), []byte("&lt;"))
-			changing = bytes.ReplaceAll(changing, []byte(">"), []byte("&gt;"))
-
-			reps = append(reps, repl{
-				start: seg.Start,
-				stop:  seg.Stop,
-				value: changing,
-			})
-		}
-
+		reps = handleRawHTML(n, source, reps)
+		reps = handleHTMLBlock(n, source, reps)
 		return ast.WalkContinue, nil
 	})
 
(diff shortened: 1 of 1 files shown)


gen-compdocs/generators/doc.go
```diff
@@ -506,3 +511,57 @@ func processUsage(usage string) string {
...
+
+	// iterating node
+	_ = ast.Walk(doc, func(n ast.Node, entering bool) (ast.WalkStatus, error) {
+		if !entering {
+			return ast.WalkContinue, nil
+		}
+
+		// when pair of '<', '>' characters found in the text and it is treated as a RawHTML node,
+		// then escape the text in RawHTML nodes.
+		t, ok := n.(*ast.RawHTML)
```
tengqm: I'm not sure how we decide if a piece of text is raw HTML.

ingyeoking13: @tengqm 
Hi,

I tried several approaches to fix linked issue and found that `goldmark` library already has a parser for recognizing raw HTML, `goldmark/parser/raw_html.go`.
goldmark has also parser for `code span` and `code block`.

I struggled to choose between 
1. after escaping '<' to '\&lt;' (like before), unescaping '\&lt;' to '<' in specific syntax(code block, code span) or 
2. escaping '<' only when goldmark parsed it as rawHTML. 
  
I think the latter way is simpler.

Caesarsage:
This handles inline RawHTML, but Goldmark parses block-level HTML as ast.HTMLBlock. For example, <div>\ncontent\n</div> and <script>alert(1)</script> remain unchanged, whereas the previous implementation escaped them.

ingyeoking13:
@Caesarsage 
> This handles inline RawHTML, but Goldmark parses block-level HTML as ast.HTMLBlock. For example,
> 
> \ncontent\n
> and <script>alert(1)</script> remain unchanged, whereas the previous implementation escaped them.

Thank you for review! :)
I missed that case, and I'll edit code soon.

tengqm:
@ingyeoking13 Please sign the CLA.

Regarding the "hacks" (I hate to use this word, but you'll know what I mean), I have some different thoughts. The reason we are doing these escaping and/or unescaping is that the raw content we use is extracted from the upstream source code. I saw many stars (`*`) in the upstream source comment, such as `*.k8s.io`. It is readable to human as Go comment, but these stars will confuse the markdown parser. I also witnessed some misused `<` and `>` characters in the comments, bad-formatted lists that are difficult to parse in markdown ... so on and so forth.

The question then is whether we should strive to be as smart as possible, to be as tolerant as possible, or we should fix the bad comments in the source code? 

Some examples:

- bad embedded yaml:
  * source: https://github.com/kubernetes/kubelet/blob/master/config/v1beta1/types.go#L224-L231
  * image
- misused `<` or `>`:
  * https://github.com/kubernetes/apiserver/blob/master/pkg/apis/apiserver/v1beta1/types.go#L360-L365
  *  image
- misused stars:
  * https://github.com/kubernetes/kubelet/blob/master/config/v1beta1/types.go#L1112-L1128
  * image
- bad paragraphs:
  * https://github.com/kubernetes/api/blob/master/admissionregistration/v1/types.go#L311-L350
  *  image

There are many more examples of these -- developers write comments freely, we struggle to generate readable reference docs. "Hacking" the generator is a dead end. We have to solve the problem by fixing the comments. Make sure the comments are properly formatted:

- use markdown indentation when appropriate;
- always use JSON field names rather than Go field names;
- use backticks when there are special characters;
- encapsulate code snippets with triple backtiqs;
- use starts, ampersands, angle brackets carefully;
- ...

The problem cannot be cleanly fixed at the generator end.

ingyeoking13:
@tengqm

Hi, thank you for your comments.

I now understand that there are many formatting issues like the cases you mentioned. When I first encountered this problem, I thought fixing the upstream source comments would be an inverted approach. After reading your explanation, I can see why improving the upstream comments is a better long-term solution.

Thank you for sharing your insight.

Do you think this PR is still good to go, or should I close it?

Regarding the CLA issue, I found that it was caused by my Git author email. I'll fix it and update you afterwards.

ingyeoking13:
@tengqm It looks like the CLA check has passed now

ingyeoking13:
close this PR.

ingyeoking13:
I think my code can be easily bug (by human error/future library change) and can be critical issue by script injection.  
I agree with @tengqm insight and I think I can contribute other way, thank you!

Issue #440, opened 2026-04-24
Emit API reference as structured Hugo shortcodes instead of HTML tables


Caesarsage:
## Motivation

The current gen-apidocs markdown backend emits API reference pages as HTML tables. This works, but couples presentation to content.

A shortcode-based approach would separate structured data (field metadata) from presentation (template rendering):

```
{{< api-reference-subfield >}}
---
apiKind: Deployment
topLevelFieldName: spec
childFieldName: selector
fieldType: LabelSelector
fieldRequired: true
---
Field description here.
{{< /api-reference-subfield >}}
```

Benefits:
- Structured data accessible to Hugo templates (enables different rendering modes from the same source)
- Presentation changes via one template, not N pages
- Responsive layouts per device via Hugo template logic

## Prerequisites

- New shortcodes defined in k/website: api-reference-toplevel, api-reference-description, api-reference-subfield, 
- gen-apidocs markdown backend updated to emit these instead of <table>
- Cross-repo coordination (design RFC → shortcodes land in k/website → gen-apidocs output updated)

## Related

- Follow-up to https://github.com/kubernetes-sigs/reference-docs/issues/434 completion (markdown backend consolidation): This will give direction
- Suggested by @lmktfy  https://github.com/kubernetes-sigs/reference-docs/pull/435#issuecomment-4306610705

Phase 1 is up in #458 (response code tables → markdown + class attribute, opt-in via `--backend=hugo-md`). Remaining phases under the same flag:

| # | Scope (working plan) |
|---|---|
| 2 | Query / path / body parameter tables → markdown + class |
| 3 | Operations wrapped in `{{< api-reference-http-request >}}` paired shortcode |
| 4 | Split resource-scoped vs collection-scoped operations; add `{{< api-reference-common-definition >}}` for shared types like `Status` |
| 5 | Field tables → markdown + `{{< api-field >}}` per cell + `{{< api-field-text >}}` wrapper |
| 6 | Heading restructure (`## Node object API type {#resource}`, `` ## `.spec` {#NodeSpec} ``, and similar) |
| 7 | Front matter additions: `api_metadata.apiGroup`, `code_import.go` |
| 8 | Preamble shortcodes (`api-object-preamble`, `api-collection-preamble`, `api-type-preamble`, `api-operations-preamble`) |

[...]

PR #459, opened 2026-06-21
refactor: split template helpers and slug utils out of markdown.go

Caesarsage:
Pure relocation; no behavior change. Tests pass without golden updates.

- New `template_helpers.go` holds the `resource.tmpl` parse block + `escape`, `hugoRef`, and their regex vars.
- Slug helpers (`anchor`, `kebabCase`, `kebabName`, `groupVersionString`, `operationSlug`, `constValueFor`) move into the existing `util.go`.
- `markdown.go` drops the moved symbols + now-unused imports.

`markdown.go` had grown to ~950 lines mixing view-model, writer methods, slug helpers, and template primitives. The extraction lets each file have one concern. 

Sets up cleaner ground for the upcoming #440 PRs.

### DIFF
--- a/gen-apidocs/generators/util.go
+++ b/gen-apidocs/generators/util.go
@@ -18,11 +18,63 @@ package generators
 
 import (
 	"fmt"
-
+	"regexp"
 	"strings"
 
 	"github.com/kubernetes-sigs/reference-docs/gen-apidocs/generators/api"
 )
+
+var (
+	anchorRegex    = regexp.MustCompile(`[^a-zA-Z0-9]+`)
+	kebabBoundary1 = regexp.MustCompile(`([a-z0-9])([A-Z])`)
+	kebabBoundary2 = regexp.MustCompile(`([A-Z])([A-Z][a-z])`)
+)
+
+// anchor produces a stable in-page anchor from a kind or section name.
+func anchor(s string) string {
+	return strings.Trim(anchorRegex.ReplaceAllString(s, "-"), "-")
+}
+
+// kebabCase lowercases and slugifies any string into a kebab-case identifier
+// used for directory and file names.
+func kebabCase(s string) string {
+	return strings.Trim(anchorRegex.ReplaceAllString(strings.ToLower(s), "-"), "-")
+}
+
+// kebabName converts a CamelCase API kind (e.g. "PodTemplate") into its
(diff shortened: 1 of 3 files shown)


gen-apidocs/generators/util.go
```diff
@@ -18,12 +18,64 @@ package generators
...
+
+// anchor produces a stable in-page anchor from a kind or section name.
+func anchor(s string) string {
+	return strings.Trim(anchorRegex.ReplaceAllString(s, "-"), "-")
+}
+
+// kebabCase lowercases and slugifies any string into a kebab-case identifier
+// used for directory and file names.
+func kebabCase(s string) string {
+	return strings.Trim(anchorRegex.ReplaceAllString(strings.ToLower(s), "-"), "-")
```
lavishpal, line 41: `operationSlug` and `kebabCase` have identical implementations. is keeping them separate intentional to allow independent divergence later, or should `operationSlug` just assign to `kebabCase`

lavishpal:
Only wanna understand your intention that why is `template_helpers.go` a separate file instead of merging into `util.go` and  also `escape(`) is in template_helpers.go but it's a pure string transform, not template-specific. The template FuncMap could reference it from `util.go`

Caesarsage:
> Only wanna understand your intention that why is `template_helpers.go` a separate file instead of merging into `util.go` and  also `escape(`) is in template_helpers.go but it's a pure string transform, not template-specific. The template FuncMap could reference it from `util.go`



> Only wanna understand your intention that why is `template_helpers.go` a separate file instead of merging into `util.go` and  also `escape(`) is in template_helpers.go but it's a pure string transform, not template-specific. The template FuncMap could reference it from `util.go`

Two design choices, each with trade-offs:

**On the separate file:** `template_helpers.go` holds the template specific functions registers. They're tightly coupled, anyone wanting to know "what helpers does the template have access to" reads one file end-to-end. Folding into `util.go` would mix template parsing with the existing `PrintInfo` diagnostic and slug helpers, three unrelated concerns in one file.

**On `escape()` placement:** you're right that it's a pure string transform. So are `hugoRef`, `mdCell`, `hugoShortcode`, and `dict`, all pure functions whose only caller is the FuncMap. I grouped them with the parse block so the boundary is obvious: "these exist to support the template, reached by these registered names." Moving `escape` alone to `util.go` would split the group across files.

lavishpal:
Make sense. 
/lgtm

PR #461, opened 2026-06-25
fix: correct description for ClusterTrustBundle in swagger.json

adityajoshi12:
Fix for https://github.com/kubernetes/website/issues/56221


Caesarsage:
Thanks you @adityajoshi12 

This file is generated from upstream k/k so that where the correct fix will be not here

lmktfy (commented):
This is the wrong place to fix the change. In this repository we intend only to have the generator code committed. Right now some generated files **are** committed, but we have an ambition to fix that.

Instead, the fix will happen within https://github.com/kubernetes/kubernetes

You could file an issue there.

/close

PR #472, opened 2026-08-26
Updates the configuration API reference for 1.37, and fixes two things that kept pages from reaching the website.

Caesarsage:
Publish the metrics pages under external-api and regenerate on copy
- `copyconfigapi` copied every generated page into config-api/, including the
  four metrics pages that the website publishes under external-api/ and five
  pages it does not publish at all
- `configapi` ran genref's default target, which only builds the binary, so the
  copy could ship stale output

Point genref at Kubernetes v1.37.0
- k8s.io/* staging requirements v0.36.0 -> v0.37.0
- externalPackages link target now points at the v1.37 API reference
- adds Kubernetes Metrics (v1), served alongside v1beta1

Emit front matter for API packages without a group
- genref only emitted front matter when a package belonged to an API group, so
  pages for groupless packages had no title and were never published
- packages parsed from an API's configured path are now marked primary and own
  the page; packages pulled in through includes cannot
- the version alone is used as `package` when there is no group, which removes
  the hardcoded kubeconfig case
- kubelet-stats.v1alpha1 leaves the skip list, since it now generates correctly

Three pages gain front matter: kubelet-stats.v1alpha1,
cloud-provider-service.v1alpha1, kube-scheduler-extender.v1. kubeconfig.v1 gains
a blank line.

ref: https://github.com/kubernetes/website/pull/57209


Caesarsage:
A couple of things I noticed while working on this, worth a follow-up PR if we agree:

- genref/output/html/ looks dead. Nothing regenerates it anymore. should we delete and gitignore?

- genref/output/md/ is tracked, while the two other generators (gen-apidocs/build/ and gen-compdocs/build/)
  are ignored. The same content usually lands in a matching k/website PR anyway.
  Is the tracked copy something we need for reviewing config API changes, or
  would we rather it be ignored too?

Happy to do either, just wanted to check before touching it.

dipesh-rawat (commented):
/lgtm

PR #473, opened 2026-09-04
gen-apidocs: include create-only resources in the generated ToC

### LINKED ISSUE #57313: Authentication v1 API reference has disappeared from navigation as of 1.36
**Problem:**

Prior to 1.36, there was an authentication API reference, including information on the TokenRequest API:

https://v1-35.docs.kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/

As of 1.36, this page now redirects to a completely irrelevant page in the storage CSI docs:

https://v1-36.docs.kubernetes.io/docs/reference/kubernetes-api/storage/csi-driver-v1/#TokenRequest

However, there are still links to this page elsewhere in the documentation:

https://v1-36.docs.kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#tokenrequest-api

And there's no indication that the token request API is deprecated or has been removed.

**Proposed Solution:**

Put the authentation v1 reference docs back (and I'm guessing, the authorization v1 docs also need to be replaced). There is nowhere else in the docs (that I've been able to find) that describe how to use this API.

abhinav-phi:
#### Problem

With `--auto-detect`, `IsTopLevelResource` only recognized definitions that have a **List** operation ("Read Operations" category, `List`/`List All Namespaces`). Resources that only support **Create** were therefore never classified as top-level, and `buildGroupBasedCategories` dropped them entirely — their API group's category/directory and per-resource pages were never emitted.

This silently dropped two whole groups from the generated navigation (kubernetes/website#57313):

- `authentication.k8s.io`: TokenRequest, TokenReview, SelfSubjectReview
- `authorization.k8s.io` review types: SubjectAccessReview, LocalSubjectAccessReview, SelfSubjectAccessReview, SelfSubjectRulesReview

Both groups are listed on the generated `group-versions.md` page but had no pages. Up to the v1.35 docs these pages existed (e.g. `/docs/reference/kubernetes-api/authentication-resources/token-request-v1/`); the migration to the markdown backend removed them because the generator no longer produced them. Verified against the v1.36 spec: baseline run produces no `authentication/` or `authorization/` directory.

#### Fix

1. `IsTopLevelResource` also accepts a top-level **Create** operation ("Write Operations" category, `Create` type) as evidence of a top-level resource.
2. In the `TokenRequest` special case (its operation is `createCoreV1NamespacedServiceAccountToken`, mapped in `mapOperationsToDefinitions`), the category was constructed and the operation linked, but the category was never appended to `d.OperationCategories`, and `o.Type` was never set — so even with fix 1 alone TokenRequest would still fail the check. Attach the category (with the typed operation) to the definition.

#### Result

With `--auto-detect --backend=hugo-md` on the v1.36 spec:

- `authentication/`: `_index.md`, `token-request-v1.md`, `token-review-v1.md`, `self-subject-review-v1.md`
- `authorization/`: `_index.md`, `subject-access-review-v1.md`, `local-subject-access-review-v1.md`, `self-subject-access-review-v1.md`, `self-subject-rules-review-v1.md`

matching the per-group structure of the other groups. The group `_index`/weights of existing groups shift by two steps (Authentication and Authorization now occupy their alphabetical slots); all page content of existing resources is unchanged.

#### Notes

- `TestWriteOperationGolden` / `TestWriteOperationGoldenHugoMode` pass identically with and without this change (verified locally against the merge-base and the PR head), so no golden refresh is needed in this PR.
- Downstream effect for kubernetes/website: after this lands, `make apimd-hugo && copyapimd` restores the missing group directories. The new resource pages for `authentication`/`authorization` land in `content/en/docs/reference/kubernetes-api/authentication/` and `.../authorization/`.
- Fixes kubernetes/website#57313 (together with a website-side commit syncing the regenerated content, coming separately).


gen-apidocs/generators/api/definition.go
```diff
@@ -337,6 +337,8 @@ func (s *Definitions) FindNewestVersion(group, kind string) string {
 // IsTopLevelResource returns true if this definition represents a top-level
 // API resource (one that has its own List endpoint), as opposed to a
 // subresource (Scale, Eviction) or utility type (Status, WatchEvent).
+// Some resources, such as TokenRequest and the SubjectAccessReview family,
+// only support Create; they are top-level resources too.
```
Caesarsage, line 341: The first sentence is now the opposite of what the function does, and Binding (a pods/binding subresource) gets promoted by this change, so the "as opposed to a subresource" part doesn't hold either.

```suggestion
// IsTopLevelResource reports whether this definition is a top-level API
// resource, as opposed to a utility type (Status, WatchEvent) or a field type
// that never appears on the wire by itself.
//
// A List endpoint is the usual marker, but create-only resources such as
// TokenRequest and the SubjectAccessReview family qualify too.
```

gen-apidocs/generators/api/definition.go
```diff
@@ -351,6 +353,19 @@ func (d *Definition) IsTopLevelResource() bool {
...
+		}
+		for _, op := range c.Operations {
+			if op == nil {
+				continue
+			}
+			if op.Type.Name == "Create" {
+				return true
+			}
+		}
+	}
```
Caesarsage, line 368:
```suggestion

```

There is no need for this extra loop. 

This can be one pass instead of walking OperationCategories twice, same conditions, same result:


```
for _, c := range d.OperationCategories {
		if c == nil {
			continue
		}
		for _, op := range c.Operations {
			if op == nil {
				continue
			}
			switch c.Name {
			case "Read Operations":
				if op.Type.Name == "List" || op.Type.Name == "List All Namespaces" {
					return true
				}
			case "Write Operations":
				if op.Type.Name == "Create" {
					return true
				}
			}
		}
	}
	return false

```

gen-apidocs/generators/api/config.go
```diff
@@ -831,10 +831,14 @@ func (c *Config) mapOperationsToDefinitions() error {
 
 				o.Definition = d
 				o.Definition.InToc = true
```
Caesarsage, line 833: Now that the category is attached and buildGroupBasedCategories marks the definition, this line is redundant  and it's the other half of what broke TokenRequest in the first place: it told writeDefinitions to skip the page while the missing category kept it out of the ToC, so it fell through both. Worth dropping so the fix reads on its own.

gen-apidocs/generators/api/config.go
Caesarsage, line 819: nit: Since you're here, this XXX: reads like a bug marker but it's just describing the operation ID. Something like "TokenRequest is created through a ServiceAccount subresource, so its operation ID doesn't follow the ${group}${version}${resource} pattern the generic matcher below uses" would save the next person a detour

Caesarsage (commented):
Nice find, the dropped OperationCategory is easy to miss. I reproduced your result on the v1.37 spec: both groups come back, 162 pages either way.

  Have you copied the output into k/website and built it locally?

https://deploy-preview-57160--kubernetes-io-main-staging.netlify.app/docs/contribute/generate-ref-docs/release-generation/

Mine failed with 18 REF_NOT_FOUND errors. Three of the admission-policy pages relref the definitions/ pages you're removing, so the website change needs to go in at the same time as this one.

While you're there, look at the page you just created, authentication/token-request-v1.md. Its request body and all three responses link to the CSI driver page instead of to itself. There are two different TokenRequest kinds, authentication.v1 and storage.v1, and the link map in markdown.go is keyed on the kind name alone, so one overwrites the other. Same thing makes Pod's eviction op point at the wrong Eviction today. So this needs a fix in markdown.go as well. That can be a follow-up PR, but please hold this one until it lands.

Last thing: the golden tests pass on your base. That golden came in with #460, which you're already on, so that note in the description can go.

Left a few smaller things inline.

abhinav-phi:
Thanks @Caesarsage for the quick review and for reproducing it on the v1.37 spec — all four suggestions are applied in 9cecf0f:

- `IsTopLevelResource` is now a single pass with a switch, and the doc comment no longer claims List-only resources are excluded (I re-verified the regenerated output after the rework: `authentication/` and `authorization/` still come out with the same pages, and Binding is still promoted to a resource page, consistent with the v1.35 docs where `workload-resources/binding-v1` existed).
- Dropped the redundant `o.Definition.InToc = true` — agreed, attaching the category is what places the definition, and the stray flag was exactly what made TokenRequest fall through both writeDefinitions and the ToC before.
- Reworded the `XXX:` comment to explain the operation-ID mismatch.

To answer your question: I did **not** copy the output into k/website and build it there. Two reasons:

1. **Format drift between reference-docs master and the site.** The content committed on kubernetes/website main still matches the hugo-md output as of the backend switch (May 13): `import:` string front matter, HTML tables, no preamble shortcodes. Master's output today (after 31032f6) emits `code_import:`, `{{< api-object-preamble >}}` and friends — and none of those shortcodes are defined in kubernetes/website's layouts (or docsy), so pasting master-format pages today would break the Hugo build. I checked this while preparing the link-repair PR (kubernetes/website#57414), which fixes the 16 stranded inbound links in the meantime; the full page restore should ride the next scheduled regen once the shortcode templates land on the site side (issue #440 territory).

2. **A pre-existing nondeterminism I hit while diffing runs** (not from this change — reproduces on master without it, verified by stashing): two consecutive runs of `--backend=hugo-md` can differ in the `core` vs `events` category weights (130/140 swap in `_index.md` front matter) and in the ref-link disambiguation of the colliding `event-v1.md` filenames (`{{< ref "event-v1#Event" >}}` vs `{{< ref "../events/event-v1#Event" >}}` in `core/event-v1.md`). Looks like a map-iteration-order dependency in the link map / weight assignment when two groups have same-named filenames. Worth a follow-up issue since it will make every website regen PR noisy — happy to file and take a stab at it if you agree.

abhinav-phi:
@Caesarsage all four of your suggestions are applied in 9cecf0f (single-pass `switch` + corrected doc comment, redundant `InToc` dropped, `XXX:` comment reworded). I re-ran `--backend=hugo-md` on the v1.36 spec after the rework — `authentication/` and `authorization/` still emit with the same page set as before, so behavior is unchanged.

Could I ask for a re-review when you have a moment? Happy to iterate on anything else.

abhinav-phi:
Hi @Caesarsage, a consolidated update since the re-review request, with one correction from my side.

First, the golden-test caveat in the PR notes was inaccurate as worded, so I have corrected it in the description rather than leaving it. I ran `TestWriteOperationGolden` and `TestWriteOperationGoldenHugoMode` locally against both the merge-base and this PR head: both pass identically in both states, so no golden refresh is needed here. Whatever failure I saw earlier does not reproduce against current master, and I did not want a stale caveat confusing reviewers.

Second, two items from my earlier note still sit outside this PR. The colliding `event-v1.md` ref-link nondeterminism I reported has no follow-up issue and no owner; I offered to file and attempt it, but only if you agree, so please tell me whether you want that filed (by me or someone else) or left alone. On the website side, the link-repair PR (kubernetes/website#57414) and the shortcode-templates track (issue #440) are both still open, so the coordinated regen and any website build validation remain blocked behind those and cannot be completed from here.

Nothing in the code changed with this update. Could I ask for a re-review when you have a moment? Happy to iterate on anything else.

